Issue #7, opened 2018-06-06
Allow import of inner modules


jeanduplessis:
First off thank you for taking the effort to write this addon.

I'm trying to move from using ember-browserify, and all my imports work except Firebase.

package.json:
```
"devDependencies": {
  ...
  "firebase": "5.0.4",
  ...
}
```

in my code as per Firebase's instruction here: https://www.npmjs.com/package/firebase#include-only-the-features-you-need 
```javascript
import firebase from 'firebase/app';

// These imports load individual services into the firebase namespace.
import 'firebase/messaging';
```

When I run the app I get the following error: `Error: Could not find module 'firebase/app' imported from 'samewave-client/services/browser-push-notifications'`.

ef4:
We need to add support for importing inner modules from within packages.

In practice it's just necessary because node supports it, so people expect it to work, so library authors design APIs that need it.

(In a perfect world, I would rather we didn't support this, because it's better to have package authors explicitly expose their public API via the package's main entrypoint. But that ship has sailed.)

jeanduplessis:
Not sure this is an ember-auto-import issue anymore but as an FYI the importing of firebase/app and firebase/messaging didn't work with 0.2.2 for me. 
It doesn't add the messaging function to the firebase app prototype. 

I can see it defines the modules in vendor.js for firebase/app and firebase/messaging.

ef4:
If you can share a reproduction I will be happy to take a look.

jeanduplessis:
@ef4 thanks, I'll create a new ember project later that showcases the issue and submit as a new issue.

PR #10, opened 2018-06-08
Allow import of inner modules

jamesdaniels:
Hey there, Firebase team member here; I was working on updating emberfire and considering using this project when I saw #7; I took a quick pass on a solution.

Thoughts?

### DIFF
--- a/lib/bundler.js
+++ b/lib/bundler.js
@@ -14,7 +14,7 @@ module.exports = class Bundler extends Plugin {
   }
 
   build() {
-    let dependencyTargets = this._analyzer.targets.filter(t => this._depFinder.hasDependency(t) && !this._depFinder.isEmberAddon(t));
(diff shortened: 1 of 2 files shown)


ef4:
Thanks. The name-handling logic here is helpful, but there are some other pieces I want to land before I merge this.

Once we allow things other than package mains to be rollup entrypoints, our odds of rolling up duplicates go up a lot. We'll want to pass (at least) all the modules from the same package to rollup in a single step so it can deduplicate ones that already call each other.

jamesdaniels:
Sounds good, in the meantime I'll work on the package naming bit to detect N-levels (rather than just two) and add a couple tests if that is helpful.

ef4:
Closing in favor of #16, this should be all set now. Thanks.

PR #34, opened 2018-06-21
Support ember-cli-typescript

jamescdavis:
`ember-auto-import` currently does not work correctly with [`ember-cli-typescript`](https://github.com/typed-ember/ember-cli-typescript) because it runs before `ember-cli-typescript` has compiled .ts files to .js. This PR forces `ember-auto-import` to run after `ember-cli-typescript` so  that the analyzer finds the compiled .js files in the tree.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,9 @@
     "configPath": "tests/dummy/config",
     "before": [
       "ember-cli-babel"
+    ],
+    "after": [
+      "ember-cli-typescript"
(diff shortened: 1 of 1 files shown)


ef4:
Excellent, thanks. 

Presumably ember-cli-typescript is configured to leave imports alone so that ember-cli-babel will deal with them. As long as that remains true this will work.

PR #44, opened 2018-06-29
fixing paths in Window OS

houfeng0923:
hi  @ef4 : 

when we use this in window,  building is failed  : 

```
ERROR in ./tmp/ember_auto_import_webpack-staging_dir-EqQ18vBa.tmp/entry.js 10:136
Module parse failed: Bad character escape sequence (10:136)
You may need an appropriate loader to handle this file type.
|     window.define('numeral', [], function() { return require('E:\work\js\node_modules\numeral\numeral.js'); });
>     window.define('node-uuid', [], function() { return require('E:\work\js\node_modules\node-uuid\uuid.js'); });

```

so, i  pull this request for you  to fix the path problem .

### DIFF
--- a/ts/webpack.ts
+++ b/ts/webpack.ts
@@ -41,8 +41,12 @@ export default class WebpackBundler {
   }
 
   private writeEntryFile(modules){
-    let moduleList = Object.keys(modules).map(specifier => ({ specifier, entrypoint: modules[specifier].entrypoint}));
+    let moduleList = Object.keys(modules).map(specifier => ({ specifier, entrypoint: this.normalizePath(modules[specifier].entrypoint) }));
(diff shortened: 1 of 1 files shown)


benmurden:
Tried this with Node's `path.normalize` locally and I can confirm that it does **not** work - the path is already normalized, something further along is treating the backslashes as escape characters.

The proposed fix does work for me.

houfeng0923:
hi @ef4 ,   I did it yet . Output of `normalize` is the same as input , so i think we need resolve it  in this direct way .  :-)

ef4:
Thanks, I see now what was happening. This is fixed in 1.0.1.

PR #57, opened 2018-07-11
Lazy loading via dynamic import

ef4:
This is nearly complete. Remaining blockers:

 - [x] make dynamic import work in fastboot
 - [x] profile the performance impact of the deeper AST traversal on some big apps, and maybe optimize it more.

### DIFF
--- a/ts/auto-import.ts
+++ b/ts/auto-import.ts
@@ -43,7 +43,7 @@ export default class AutoImport{
     analyze(tree, appOrAddon){
         let pack = Package.lookup(appOrAddon);
         this.packages.add(pack);
-        let analyzer = new Analyzer(debugTree(tree, `preprocessor:input`), pack);
+        let analyzer = new Analyzer(debugTree(tree, `preprocessor:input-${this.analyzers.size}`), pack);
         this.analyzers.set(analyzer, pack);
         this.bundlers.forEach(bundler => bundler.unsafeConnect(analyzer));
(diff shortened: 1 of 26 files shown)


ts/auto-import.ts
```diff
@@ -52,40 +54,40 @@ export default class AutoImport{
...
-        return [appBundler, testsBundler];
+    treeForVendor(){
+      return this.makeTree();
     }
 
-    treeForVendor(tree){
-        return new MergeTrees([tree].concat(this.bundlers.map(b => b.tree)).filter(Boolean));
+    treeForPublic() {
+      return debugTree(new Funnel(this.makeTree(), {
+        srcDir: 'ember-auto-import',
```
josemarluedke: I wonder if the folder name could be something more generic instead of `ember-auto-import`, as this will be part of the url for these files.  Maybe something like `chunks`?

ef4: `srcDir` is where `Funnel` reads from, not where it writes to. At the moment I'm storing the chunks directly in the `assets` folder, named like `chunk.0.js`, `chunk.1.js`.

ef4:
No detectable performance impact when tested against travis-web. Since we were already doing a full parse, just traversing the AST to look for dynamic imports isn't a significant new cost.

knownasilya:
🚀

ef4:
Released in 1.2.0.

Issue #58, opened 2018-07-11
Error during eval. Possible bad escaping of module?


cibernox:
This happens when importing `luxon` (https://moment.github.io/luxon/) on an app or addon.

Steps to reproduce:

1. Create a new app or addon
2. Add "ember-auto-import" and "luxon" as dependencies.
3. In some file that is requires immediately, like the application controller, add `import { DateTime } from "luxon"`
4. The error I get is: 
```
Uncaught SyntaxError: Unexpected token :
    at Object../node_modules/luxon/build/cjs-browser/luxon.js
```

I created a repro to help: https://github.com/cibernox/addon-with-luxon (it's an addon it can also be reproduced on an app)

Since I don't think luxon really has a syntax error in its source code, I suspect that the escaping to convert the library to a string has a bug.

cibernox:
Same. I had this working but now it fails. I tried downgrading luxon and ember-auto-import but it keeps failing. I'm almost sure that it's failing because of an update in webpack.
(use Ember they said, you won't have to worry about webpack they said...)

After further digging I discovered that the syntax error is in the very end of the eval'ed string, in the webpack comment: 
```
exports.FixedOffsetZone = FixedOffsetZone;
exports.IANAZone = IANAZone;
exports.LocalZone = LocalZone;
exports.Settings = Settings;
sourceURL=webpack://__ember_auto_import__/./node_modules/luxon/build/cjs-browser/luxon.js?
```

Seems that the last line should be a comment but it's not.

ef4:
I think this is a bug in broccoli-concat. Here's the very end of the module's definition before and after concat:

```
exports.Settings = Settings;\n//# sourceMappingURL=luxon.js.map\n\n\n//# sourceURL=webpack://__ember_auto_import__/./node_modules/luxon/build/cjs-browser/luxon.js?");
exports.Settings = Settings;\nsourceURL=webpack://__ember_auto_import__/./node_modules/luxon/build/cjs-browser/luxon.js?");
```

ef4:
@cibernox's repro app stops being broken if you disable sourcemaps:

```
diff --git a/ember-cli-build.js b/ember-cli-build.js
index dc5a39e..89729fe 100644
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -5,6 +5,9 @@ const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 module.exports = function(defaults) {
   let app = new EmberAddon(defaults, {
[...]

PR #60, opened 2018-07-12
Add debugging of output via broccoli-debug

rwjblue:
This allows easier debugging of the webpack build output for both app and test bundles.

When using:

```
BROCCOLI_DEBUG=ember-auto-import:* ember build
```

You will now see both the `DEBUG/ember-auto-import/preprocessor/input` files (which were previously available) and `DEBUG/ember-auto-import/vendor/output/` files.

---

This _should_ help isolate which packages are responsible for the issue in https://github.com/ef4/ember-auto-import/issues/58 (e.g. is webpack just emitting invalid values, is something later in the pipeline mucking up the sourcemap comment, etc...

### DIFF
--- a/ts/auto-import.ts
+++ b/ts/auto-import.ts
@@ -86,6 +86,8 @@ export default class AutoImport{
     }
 
     treeForVendor(tree){
-        return new MergeTrees([tree].concat(this.bundlers.map(b => b.tree)).filter(Boolean));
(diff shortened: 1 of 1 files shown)


ef4:
Thanks. I added the same one in my dynamic-import branch, which is nearly ready to merge.

PR #61, opened 2018-07-12
fix rebuilds

ef4:
It was not correct to unlink the webpack build output, because in development we rely on having a persistent webpack that does incremental rebuilds, and it doesn't re-output the files we unlink.


ef4:
Our test coverage for initial builds is quite good, but not so good for rebuilds!

PR #67, opened 2018-07-16
allow multiple copies of same package version

### LINKED ISSUE #65: Broccoli Plugin: BundlerPlugin build error
Hey there,

I'm attempting to pull a dependency in from an add-on. That dependency (leaflet.js) is also a dependency of the app consuming the add-on. I believe I have my ducks in a row with ember-auto-import as well as leaflet in dependencies and not devDependencies of the add-on. My add-on code is being imported from the /addon directory, but I have imported the leaflet dependency in /app as well:

`import Leaflet from 'leaflet';`
`export { default } from 'imi/components/map-config'; `

ember-auto-import is a devDependency in the consuming app while leaflet is a dependency. 

This code organization is working seamlessly with other addon/app combinations. Leaflet seems to be a problem. I am using leaflet version 1.3.1 which is an ES6 module. I am confused by the following build error: 

 Error: imi and imi-application are using different versions of leaflet (/imi/node_modules/leaflet/dist/leaflet-src.js vs /imi-application/node_modules/leaflet/dist/leaflet-src.js)
    at Splitter.<anonymous> (/imi-application/node_modules/ember-auto-import/js/splitter.js:77:31)
    at Generator.next (<anonymous>)
    at fulfilled (/imi-application/node_modules/ember-auto-import/js/splitter.js:4:58) 

Thanks in advance for clarification on this issue.

ef4:
Closes #65 .


ef4:
To be clear, this doesn't cause duplicate copies to be included in an app. It allows one copy to stand in for all the others, as long as those others have the same version number.

PR #75, opened 2018-07-26
Configure Webpack's `publicPath` correctly

### LINKED ISSUE #64: Dynamic import in dependency references wrong location
I'm trying to leverage `ember-auto-import` in one of my addons ([`ember-cli-stencil`](https://github.com/alexlafroscia/ember-cli-stencil)) to solve an issue where [Stencil](https://stenciljs.com/) packages [need to be imported through ESM](https://github.com/alexlafroscia/ember-cli-stencil/issues/5).

The packages contain dynamic imports to lazy-load their component definitions. This seems to be OK, although the location of the "chunk" that it's importing is wrong.T

The file is trying to be loaded at `localhost:4200/chunk.whatever.js` (which fails)

screen shot 2018-07-14 at 4 35 24 pm

when it actually lives at `localhost:4200/assets/chunk.whatever.js`

screen shot 2018-07-14 at 4 35 48 pm

alexlafroscia:
This change allows WebPack to correctly build dynamic imports found within imported dependencies themselves.

For example, if you import `foo` which has a dynamic import somewhere inside of it, Webpack might try to load some file such as `localhost:4200/chunk.abcd.js`.

`ember-auto-import` correctly generated `chunk.abcd.js`, but the location is `localhost:4200/assets/chunk.abcd.js` (following the Ember CLI convention of putting JS files in `assets/`). By telling Webpack that fact, it will dynamically import the files from the correct location.

I don't believe that the `assets` directory is something users configure, but if I'm wrong about that, we should read that configuration from the host app and use that path instead. I did some searching online and couldn't find a way to change it, though, so this should be safe as far as I know.

I came to this conclusion as the solution by finding a similar issue filed against Webpack itself and the solutions people recommended in that case.

https://github.com/webpack/webpack/issues/7417

Closes #64

### DIFF
--- a/ts/webpack.ts
+++ b/ts/webpack.ts
@@ -1,14 +1,19 @@
-import webpack from 'webpack';
-import { join } from 'path';
-import { merge } from 'lodash';
-import quickTemp from 'quick-temp';
-import { writeFileSync, readFileSync, readdirSync, emptyDirSync } from 'fs-extra';
-import { compile, registerHelper } from 'handlebars';
-import jsStringEscape from 'js-string-escape';
-import { BundleDependencies } from './splitter';
-import { BundlerHook } from './bundler';
+import webpack from "webpack";
+import { join } from "path";
+import { merge } from "lodash";
+import quickTemp from "quick-temp";
+import {
+  writeFileSync,
+  readFileSync,
+  readdirSync,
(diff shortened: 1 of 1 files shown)


ef4:
I think I see what's wrong, following up in #80.

PR #86, opened 2018-08-03
Use filter method to support Internet Explorer 11

bobisjan:


ef4:
Thanks, good call.

bobisjan:
Thanks @ef4, makes sense to publish a patch release with this?

ef4:
Yes. Published in 1.2.6.

Issue #109, opened 2018-08-21
v1.2.11 Analyzer causing build error


nightire:
Getting this error after upgrade to v1.2.11 from v1.2.10, when file changes:

```
file changed index/template.hbs
Build Error (Analyzer)

EEXIST: file already exists, symlink '/Users/nightire/Code/choice-form/os-official/tmp/broccoli_persistent_filtertemplate_compiler-output_path-U5ie8xKK.tmp/os-official/index/template.js' -> '/Users/nightire/Code/choice-form/os-official/tmp/analyzer-output_path-oU0fnTzU.tmp/os-official/index/template.js'
```

here is the full log:

```
ENV Summary:

  TIME: Tue Aug 21 2018 17:17:08 GMT+0800 (China Standard Time)
  TITLE: ember
  ARGV:
  - /usr/local/bin/node
  - /Users/nightire/Code/choice-form/os-official/node_modules/.bin/ember
  - serve
  - --ssl
  EXEC_PATH: /usr/local/bin/node
  TMPDIR: /var/folders/9_/y4f4n_cd5ddbzftt3x0rz19w0000gn/T
  SHELL: /bin/zsh
  PATH:
  - /usr/local/lib/node_modules/npm/node_modules/npm-lifecycle/node-gyp-bin
  - /Users/nightire/Code/choice-form/os-official/node_modules/.bin
  - /usr/local/bin
  - /usr/bin
  - /bin
  - /usr/sbin
  - /sbin
  - /usr/local/opt/texinfo/bin
  - /usr/local/opt/sqlite/bin
  - /usr/local/opt/openssl/bin
  - /usr/local/opt/libxml2/bin
  - /usr/local/opt/gettext/bin
  - /usr/local/opt/curl/bin
  - /Users/nightire/.cargo/bin
  - /usr/local/opt/gnu-tar/libexec/gnubin
  - /usr/local/Cellar/zplug/2.4.2/bin
  - /usr/local/sbin
  PLATFORM: darwin x64
  FREEMEM: 678449152
  TOTALMEM: 17179869184
  UPTIME: 677829
  LOADAVG: 1.8515625,2.23876953125,2.1943359375
  CPUS:
  - Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz - 4000
  - Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz - 4000
  - Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz - 4000
[...]

PR #112, opened 2018-08-22
fix analyzer rebuilds

ef4:
This was a silly typo that escaped because I wasn't testing the analyzer in rebuild scenarios.

Now I am.


ef4:
Closes #109.

PR #136, opened 2018-10-24
Swap package.json keyword rollup -> webpack

chadian:
Extremely minor, but I noticed the release notes mentioned the implementation was changed in 0.2.0. I was just doing a peak under the hood to see if rollup was used anywhere and this seemed like the only leftover reference.


ef4:
Thanks. 👍

PR #137, opened 2018-10-24
get `define` and `require` from scope w/o alerting webpack

noslouch:
Sometimes an ember app might need to close over the build app and vendor
files in order to hide any ember globals from the browser's global
scope. If you have a third-party script running that tries to use
`window.require` or `window.define` in a way that's incompatible with
ember's `loader.js` implementation, for instance.

In those cases, `window.define` won't be there unless another script has
defined it, and regardless, `ember-auto-import` is interested in the
values that ember defines.

Using `eval` here allows us to get the values defined in scope (by ember),
but also prevents webpack from noticing that we're using `define` and
`require`, which would send webpack its own behavioral path of resolving
things for the build. One of those few times when `eval` is your friend.

### DIFF
--- a/ts/webpack.ts
+++ b/ts/webpack.ts
@@ -32,8 +32,8 @@ if (typeof document !== 'undefined') {
 
 module.exports = (function(){
   var w = window;
-  var d = w.define;
-  var r = w.require;
+  var d = eval('define');
+  var r = eval('require');
   w.emberAutoImportDynamic = function(specifier) {
     return r('_eai_dyn_' + specifier);
   };


ef4:
Thanks, this is good.

I hadn't considered that we were subtly changing the semantics of how `define` and `require` are found in a normal Ember app by going straight to `window` rather than using the surrounding lexical environment.

After we chatted about this in Discord, I thought of one remaining concern: some people run with strict CSP rules that forbid eval. This would be a breaking change for them.

noslouch:
Ah, good point. I wonder if there's a way to configure webpack to ignore a particular `define` or `require` statement?

ef4:
I wonder if it tries to follow assignment statements. `d = define` might actually work by itself.

noslouch:
but the tests fail :(

ef4:
Yes, that was an oops. I thought I had the tests passing locally, but I wasn't really running the new code.

noslouch:
yeah that `compile` step was tripping me up too

ef4:
Making another attempt in https://github.com/ef4/ember-auto-import/pull/138

ef4:
Thanks, this is fixed in 1.2.14.

noslouch:
👏 👏 👏 thanks ed!

PR #138, opened 2018-10-25
Scoped amd

ef4:


noslouch:
👏🏻 Tests pass!

ef4:
And I also added tests for the strict CSP case.

PR #141, opened 2018-10-25
Bugfix: noparse test can fail

ef4:
You can get a "define cannot be used indirect" exception if our noParse rule fails to match the little loader file we're using.

That can happen when your temp directory has a different realpath than what's passed to the webpack builder.


ef4:
@burieberry I think this will fix the bug you found.

ef4:
Released in 1.2.15.

PR #157, opened 2018-11-27
feat(analyzer): add `extensions` option

### LINKED ISSUE #155: Incompatibility with ember-cli-typescript v2
With the latest release of `ember-auto-import`, we can now use ember-auto-import and ember-cli-tyepscript v2 which uses babel 7 for TypeScript. 

In my initial tests, I noticed that if a TS file is importing a module, that `ember-auto-import` should include to the bundle, it is not including anymore after v2.

This bug probably is here, but I might be wrong. Let me know if this a bug on ember-cli-typescript.

I have putted a reproduction of the error in this repo: https://github.com/josemarluedke/--ts-v2-demo

You can see that after the last commit, which adds ember-cli-typescript v2, the import does not work anymore.

buschtoens:
Fixes #155.

Adds an `extensions` option. From the `README.md`:

> `extensions`: _string[]_, defaults to the [`extensions` set for `ember-cli-babel`](https://github.com/babel/ember-cli-babel#options). All files matching this list of file extensions will be analyzed for imports.

### DIFF
--- a/packages/ember-auto-import/tsconfig.json
+++ b/packages/ember-auto-import/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "include": ["./ts/**/*.ts"],
   "compilerOptions": {
+    "lib": ["es7"],
     "target": "es6",
     "module": "commonjs",
     "declaration": true,
--- a/README.md
+++ b/README.md
@@ -109,6 +109,7 @@ Suported Options
  - `forbidEval`: _boolean_, defaults to false. We use `eval` in development by default (because that is the fastest way to provide sourcemaps). If you need to comply with a strict Content Security Policy (CSP), you can set `forbidEval: true`. You will still get sourcemaps, they will just use a slower implementation.
  - `publicAssetURL`: where to load additional dynamic javascript files from. You usually don't need to set this -- the default works for any Ember app that isn't doing something unusual.
  - `webpack`: _object_, An object that will get merged into the configuration we pass to webpack. This lets you work around quirks in underlying libraries and otherwise customize the way Webpack will assemble your dependencies.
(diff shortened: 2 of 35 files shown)


packages/ember-auto-import/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "include": ["./ts/**/*.ts"],
   "compilerOptions": {
+    "lib": ["es7"],
```
buschtoens: Was required in order to allow usage of `Array#includes`, which is supported down to Node 6.4: https://node.green/#ES2016-features-Array-prototype-includes

`Array#includes` is already used in this project, but on an `any` type, which is why `tsc` has not complained yet.

ef4:
Thanks, this looks great.

I pushed a commit fixing the first problem that CI uncovered, but there are not TS validation errors too.

ef4:
(I may drop the option from public API before releasing, because the default you've chosen is almost certainly correct and picking something different is likely to result in parse errors.)

buschtoens:
Thanks! If you like I can investigate further regarding the CI failures. I can also remove the `extensions` option from the README and add inline comments stating that this option is private, if you prefer.

buschtoens:
> but there are now TS validation errors too.

These are emitted during the build of `sample-typescript2`. I did not include the `@types/*` that would make the dummy app itself valid to save on npm packages, since these errors are inert anyways.

If you like I can:

- add the `@types/*` dependencies and potentially required custom types in the `types` dir
- dial down the warnings
- roll some "dirty" custom types that cover just the parts to make the dummy valid
- just keep it as it is

buschtoens:
@ef4 Already got a preference regarding https://github.com/ef4/ember-auto-import/pull/157#issuecomment-442184236? We could merge then. 😊

buschtoens:
If anyone else is in dire need, you can use [`@buschtoens/ember-auto-import@1.2.17-beta.1`](https://www.npmjs.com/package/@buschtoens/ember-auto-import) in the meantime.

PR #183, opened 2019-01-27
Improved docs for publicAssetURL

jrjohnson:
Added specific example of when this might be needed as well as a default
value. When I encountered my chunk.js files not being loaded from `/assets`, but from `/` instead I read these docs, but the language here convinced me I should move on to something else. I'm not sure if telling the browser to async load the scripts is 'unusual' or not, but I hope this saves someone else some time.


ef4:
Thanks, I expanded the text a little so it's not confusing if people are also moving their assets to another path.

PR #191, opened 2019-02-15
fix: change from security audit

knownasilya:
https://www.npmjs.com/advisories/755

### DIFF
--- a/packages/ember-auto-import/package.json
+++ b/packages/ember-auto-import/package.json
@@ -40,7 +40,7 @@
     "enhanced-resolve": "^4.0.0",
     "fs-extra": "^6.0.1",
     "fs-tree-diff": "^1.0.0",
-    "handlebars": "^4.0.11",
+    "handlebars": "^4.0.13",
     "js-string-escape": "^1.0.1",
     "lodash": "^4.17.10",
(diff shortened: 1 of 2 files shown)


knownasilya:
Not sure where the types are coming from..

ef4:
`./packages/sample-typescript2` is a sample TypeScript ember app that we use in tests. It depends on `@types/ember`, which depends on `@types/handlebars`.

The new `handlebars` package contains its own types, which makes it a breaking change for any app that was already using `@types/handlebars`.

https://github.com/wycats/handlebars.js/issues/1499

ef4:
I don't think your change really generated the error, CI broke as soon as they published handlebars 4.1.0, which is allowed by our `^` range and contains the new conflicting types.

The security fix is available in 4.0.13, so I think switching to `~` fixes the security issue while avoiding the new types.

ef4:
Published in 1.2.20.

nknapp:
Sorry, for the mess. I'm responsible for publishing that version. I hadn't anticipated the trouble.

Now, there is one thing I'm concerned about. Once we are through the process of removing Handlebars from DefinitelyType, 4.1 should be safe again, right? Does that mean that 4.0.13 will be broken for you then?

Issue #192, opened 2019-02-15
ember build failing at webpack error (debugging help)


efx:
After updating to  ember-cli@3.7.1, `ember build`  fails:
```
ERROR in ./tmp/broccoli-35248RkE16LEYgYVA/cache-590-bundler/staging/app.js 23:47
Module parse failed: Unexpected token (23:47)
You may need an appropriate loader to handle this file type.
|     d('moment-timezone', [], function() { return require('project/node_modules/moment-timezone/index.js'); });
|     d('raf-pool', [], function() { return require('project/node_modules/raf-pool/dist/raf-pool.umd.js'); });
>     d('_eai_dyn_docx', [], function() { return import('project/node_modules/docx/build/index.js'); });
| })();
|
cleaning up...
Build Error (Bundler)

webpack returned errors to ember-auto-import
```
We are using ember-auto-import@1.2.19. The failure comes from a dynamic import of the docx package. I believe I have the configuration specified correctly. Commenting out the [`import(`](https://github.com/ef4/ember-auto-import/blob/d034c9ff12f2770a0434a28589f5e6fa13233b4b/packages/ember-auto-import/ts/webpack.ts#L43) causes `ember build` to succeed.
I [created a fresh application with a similar setup](https://github.com/efx/poc-bug/tree/ember-auto-import-bug), and it works. This suggests a configuration bug with my main application. Do you have pointers to ember-auto-import internals to further debug this?  I wanted to ask before pouring a day into fruitless debugging.

mlenderink:
I'am getting the same error trying to build a ember-addon within a mono-repo. It works fine within its own repo.

ef4:
I think that error message is claiming that `./tmp/broccoli-35248RkE16LEYgYVA/cache-590-bundler/staging/app.js` has a syntax error. Is that true? Can you share that file?

efx:
@ef4 I will grab that snippet locally. Do you what flag / env var I use to persist the broccoli artifacts?

ef4:
There may be some BROCCOLI_DEBUG tree that contains that file, but I'm not sure. 

If you run `ember s`, the file should remain there until you exit.

efx:
@ef4 thanks, `ember s` did the trick. BROCCOLI_DEBUG didn't.
Here is the output of that chunk:
```
if (typeof document !== 'undefined') {
  __webpack_public_path__ = (function(){
    var scripts = document.querySelectorAll('script');
[...]

PR #193, opened 2019-02-22
workaround upstream npm bug

efx:
Close 192. See comment for link to upstream npm bug:
https://github.com/ef4/ember-auto-import/issues/192#issuecomment-466471583


ef4:
Thanks, published in 1.2.21.

PR #199, opened 2019-03-12
README: Update "Dynamic Import" instructions to allow parallel Babel processing

Turbo87:
Passing a file path to Babel makes it possible to parallelize the work that Babel does :)


ef4:
Thanks, good catch.

Issue #187, opened 2019-02-08
forbidEval in app's ember-cli-build.js doesn't affect ember-auto-import in dependency addons


DuBistKomisch:
Following on from #50 and #139.

We're only using `ember-auto-import` via `ember-intl`, not directly in our app. Setting `autoImport: { forbidEval: true }` in our `ember-cli-build.js` doesn't seem to prevent it from using `eval`, generating dozens of CSP errors in development.

Is it possible for the setting to flow down like this? Do we need to add `ember-auto-import` as a direct dependency too to make it work? Is this something `ember-intl` has to fix instead?

ef4:
`forbidEval` is supposed to only be controlled by apps, not addons. So it's not something to fix in ember-intl. 

You're right, we need to add the ability for apps to set this even when the app itself isn't directly using auto-import.

As a workaround, I think it should work to add ember-auto-import as a dependency to your app and then it will respect your setting.

DuBistKomisch:
Thanks, I'll try adding it as a dependency.

DuBistKomisch:
Late reply, but adding as a direct dependency does seem to work. Guess it's time to finally migrate from `ember-browserify`!

ef4:
This is dependent on https://github.com/ef4/ember-auto-import/issues/196

simonihmig:
I just had the same problem, but for me `forbidEval` had no effect even though the app is already using `ember-auto-import`. I am using `ember-cli-mirage` (which depends on e-a-i as well), and a whole lot of its imports (of `miragejs`) can be found in `test-support.js` as `eval()`! 🤔

ef4:
@simonihmig the original issue above got fixed in #213, if you're still seeing this please file a new issue.

Issue #196, opened 2019-02-26
Always check for app-level configuration


ef4:
Right now, to affect auto-import's configuration you must depend (directly) on auto-import. But sometimes you have a need to affect the configuration (as used by your dependencies) even though you don't otherwise need or want a direct dependency on auto-import.

For example, a setting like `forbidEval` only belongs at the app level, it would not be appropriate on an addon. But an app that depends on an addon that uses ember-auto-import may still want to express that it needs `forbidEval`, even though the app itself doesn't have a direct dependency on auto-import.

We should always check for configuration on the app.

stefanpenner:
does https://github.com/ef4/ember-auto-import/pull/213 close this?

ef4:
Yes, this is fixed.

PR #213, opened 2019-05-20
Honor AutoImport options from host without installing ember-auto-import

tmquinn:
This allows you to set forbidEval in ember-cli-build.js at the app level,
without having to install ember-auto-import

Fixes issue https://github.com/ef4/ember-auto-import/issues/187 & https://github.com/ef4/ember-auto-import/issues/196


tmquinn:
Looking at writing a test before making this ready for review

scripts/test.sh
```diff
@@ -20,6 +20,7 @@ cd packages/sample-babel7      && yarn test
 cd packages/sample-typescript2 && yarn test
 cd packages/sample-noconflict  && yarn test
 cd packages/sample-noparse     && yarn test
-cd packages/ember-auto-import  && yarn test
+# There are no tests in here and it's throwing an error anyway
```
ef4: I couldn't reproduce this error, and would prefer not to disable this test suite, because it creates a trap for a future developer who tries to put tests in the standard place for an addon and then they don't run.

tmquinn: I do occasionally get EENT errors about existing tmp directories in the packages, but I was able to get it to pass by cleaning, re-yarn'ing

ef4: I suspect something within broccoli is sharing the same temp directories between runs and may not be completely safe to parallelize. I haven't been able to narrow down exactly what.

tmquinn:
The issue now becomes that the EmberApp returned by `_findHost()` does not include a `parent` property, whereas the code in the Package constructor appears to be expecting one. 

Am looking further to see why, but if you are already aware of what the dynamic may be here that would be helpful.

tmquinn:
Ah, so it looks like `_findHost()` is returning `.app` instead of the object that contains `.app`

tmquinn:
If this looks good, I will add some tests...

ef4:
Yes, this looks good.

tmquinn:
Tests are in.

ef4:
Nice, the tests look good.

PR #214, opened 2019-05-29
disable size hints by default

ef4:
They're not particularly actionable most of the time, and different apps have different size budgets.


Alonski:
Awesome work Ed!

PR #219, opened 2019-06-10
Fix CSS concat separator

bendemboski:
D'oh!

When concatenating CSS, we want to use just '\n' as a separator, rather than ';\n' because CSS doesn't use semi-colons outside of rules. So only use ';\n' when concatenating Javascript.


ef4:
Thanks!

ef4:
Published in 1.4.1.

PR #225, opened 2019-06-27
[FEATURE] Babel all the things

pzuraq:
Enables babel transpilation for ember-auto-import packages. Also adds the `skipBabel` configuration option, which allows users to skip babel transpilation for specific packages.


packages/a-module-dependency/index.js
```diff
@@ -0,0 +1,15 @@
+function returnUndefined() {
+  // this should throw an error unless it's been transpiled
+  return foo;
+  let foo = 123;
```
ef4: Haha, nice, I like this.

mnoble01:
@ef4 Do you have a rough timeline of when this change will be released?

ef4:
Published in 1.5.0. 🎉

PR #227, opened 2019-07-05
README - add FAQ section for Webpack polyfills

jenweber:
We ran into this in the Ember Guides and Deprecation app after updating dependencies, and if someone yarn installed by accident (since we used an npm lockfile set to version 1.2)


ef4:
Thanks.

ef4:
Also see https://github.com/ef4/ember-auto-import/issues/228 which I filed to cover the wider issue of helping people with this.

Issue #202, opened 2019-04-02
Trouble when linking addons


Jbcampbe:
Locally whenever I npm link my addon the app build almost always fails with an error.

`Error: <host-app> and <my-addon> are using different versions of <library>`

Usually the library versions are only off by a patch bump. But when I have a lot of libraries being imported it can be very time consuming getting all of the library versions to line up between the host and addon. And life gets even more painful when I need to link 2+ addons to the host (which I do have to do frequently for work). 

Is there any way for the auto import to automatically resolve the library version difference if the two versions are semver compatible (according to the host apps package.json)?

Being able to automatically resolve these issues would dramatically improve the addon development experience.

ef4:
I agree, we can make this better.

We can tolerate the presence of multiple versions as long as they satisfy the semver requirements of all consumers.

Jbcampbe:
Is this change on the roadmap at all? As we have moved to using ember-auto-import for everything this error is becoming more and more common

ef4:
The condition that needs to change is here if somebody wants to have a go:

https://github.com/ef4/ember-auto-import/blob/e7d34659666ece3a33c822971e12720680b86c89/packages/ember-auto-import/ts/splitter.ts#L148

miguelcobain:
@ef4 what is needed to move this forward in 2021?
Is it just a matter of changing that condition to a more relaxed and semver friendly one?

ef4:
@miguelcobain I started trying to write instructions and it was harder than just doing it. 😁

miguelcobain:
Thank you! 🙌

PR #236, opened 2019-07-29
Allow patch versions of libraries to pass the safe version check

omarestrella:
For #202, just to get the ball rolling.

A concern of mine is that different users have different requirements for what would satisfy a safe version. I wasn't sure of the way to be able to allow users to configure something like this, if at all. 100% open to feedback.


ef4:
Thanks for working on this.

There's not really ambiguity on what is supposed to be a safe version. The comparison we need to make is between the stated range requirement in *every* consumer's package.json against the actual version of the package being consumed.

omarestrella:
Ah, yes, I forgot that the package.json file could be a resource here. Will be taking another look soon.

Thanks!

omarestrella:
Closing this as the original repo has disappeared. I am planning on re-opening.

PR #244, opened 2019-09-26
Update: Update vulnerable packages

### LINKED ISSUE #242: High vulnerability in dependencies
Both `handlebars` and `braces` (which is included by broccoli) have high vulnerability CVE open against them.  handlebars in particular is causing issues as it is a dependency and causing audit failures in builds.  Also `ember-try` is pulling in a vulnerable version of `lodash`

handlebars:
https://npmjs.com/advisories/1164 

braces:
https://npmjs.com/advisories/786

lodash:
https://npmjs.com/advisories/1065
https://npmjs.com/advisories/782
https://npmjs.com/advisories/577

cythrawll:
resolves #242


packages/sample-addon/package.json
```diff
@@ -33,7 +33,7 @@
     "ember-cli": "*",
     "ember-cli-dependency-checker": "*",
     "ember-cli-eslint": "*",
-    "ember-cli-htmlbars-inline-precompile": "*",
+    "ember-cli-htmlbars-inline-precompile": "~2.1.0",
```
ef4: All of these sample package's package.json's should use `*` dependencies for everything. The real versions are controlled by `packages/ember-auto-import/package.json` which should be the only one that uses real versions.

cythrawll: The problem is that it brings in `ember-cli-htmlbars-inline-precompile` 3.0. Which removes compatibility with babel 6.  This was my first approach and it just broke every test that used babel 6.

cythrawll: https://travis-ci.org/ef4/ember-auto-import/builds/590091845

This is the build where you can see the breakage it causes. The problem is that this dependency being brought in is too new, that they are too old.

cythrawll: Okay so I did end up reverting MOST of the  `ember-cli-htmlbars-inline-precompile` to `*`.

Except for the packages that were failing because of the newer API: that is `sample-direct` and `sample-addon`.  In those I loosened the package floating to `<3.0.0` in order for it to float as much as possible.

NikkiAlgarCentrica:
This is great work, thanks @cythrawll 
@ef4 hopefully you can release this soon?

PR #245, opened 2019-09-26
lazily access babel config, this ensures all other "writers" have

stefanpenner:
written before ember-auto-import reads.

We noticed this while debugging https://github.com/embroider-build/embroider/pull/325.
We also intend to do some refactoring in embroider, to provide better error messages/warnings to help users transition. 

- [x] validate (confirmed locally this mitigates the motivating issue)
- [x] tests


stefanpenner:
@ef4 I'm maybe doing something wrong, locally tests pass individually but not when run in bulk. I may circle back to try and figure out why, but I'm going to move on to the MacroConfig work we discussed in embroider.

If you have a cycle or two to land ^ or something like it, that would be great. If not, I will get back to it eventually.

ef4:
This PR is green now after re-running travis. The sporadic yarn crash is annoying but unrelated.

stefanpenner:
updated \w test, let's see if CI agrees with my local test.

stefanpenner:
👯‍♂

PR #249, opened 2019-10-14
feat(splitter): add `.ts` extension to resolver

### LINKED ISSUE #248: Splitter does not respect `resolver.extensions`
When I configure webpack to support `.ts` files, I still need to append the `.ts` file extension in the `import` statement.

```ts
import bar from 'foo/bar.ts'; // works
import bar from 'foo/bar';    // breaks
```

This is the config. It's not perfect yet and likely misses the `preset-env`, but it's sufficient to load `.ts` files directly from yarn workspaces.

```ts
{
  webpack: {
    resolve: {
      extensions: ['.ts', '.js']
    },
    module: {
      rules: [
        {
          test: /\.ts$/,
          use: {
            loader: 'babel-loader-8',
            options: {
              presets: ['@babel/preset-typescript']
            }
          }
        }
      ]
    }
  }
}
```

This can be fixed by adding `.ts` to the list of extensions in the splitter:

https://github.com/ef4/ember-auto-import/blob/e7d34659666ece3a33c822971e12720680b86c89/packages/ember-auto-import/ts/splitter.ts#L17-L22

The [docs for `enhanced-resolve`](https://github.com/webpack/enhanced-resolve#passing-options-from-webpack) actually suggest that the `webpack.resolve` options should be passed through.

Tangentially related to #74.

I am happy to send in a PR, just let me know, what you think the optimal solution is. I'd suggest passing the `webpack.resolve` through to the splitter and creating individual resolvers for each set of options, in case there are diverging options.

buschtoens:
Fixes #248.

This PR adds a new `sample-typescript` test support package that uses `ember-cli-typescript@3` without `ember-decorators`. This was not strictly necessary, but I felt that it was good to test with the latest tech.

### DIFF
--- a/packages/ember-auto-import/ts/splitter.ts
+++ b/packages/ember-auto-import/ts/splitter.ts
@@ -17,7 +17,7 @@ const debug = makeDebug('ember-auto-import:splitter');
 const resolver = ResolverFactory.createResolver({
   // upstream types seem to be broken here
   fileSystem: new CachedInputFileSystem(new NodeJsInputFileSystem(), 4000) as unknown as AbstractInputFileSystem,
-  extensions: ['.js', '.json'],
+  extensions: ['.ts', '.js', '.json'],
   mainFields: ['browser', 'module', 'main']
 });
 
--- a/packages/ember-auto-import/ts/splitter.ts
+++ b/packages/ember-auto-import/ts/splitter.ts
@@ -17,7 +17,7 @@ const debug = makeDebug('ember-auto-import:splitter');
 const resolver = ResolverFactory.createResolver({
   // upstream types seem to be broken here
   fileSystem: new CachedInputFileSystem(new NodeJsInputFileSystem(), 4000) as unknown as AbstractInputFileSystem,
-  extensions: ['.ts', '.js', '.json'],
+  extensions: ['.js', '.ts', '.json'],
   mainFields: ['browser', 'module', 'main']
 });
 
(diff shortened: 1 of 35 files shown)


packages/ember-auto-import/ts/splitter.ts
```diff
@@ -17,7 +17,7 @@ const debug = makeDebug('ember-auto-import:splitter');
 const resolver = ResolverFactory.createResolver({
   // upstream types seem to be broken here
   fileSystem: new CachedInputFileSystem(new NodeJsInputFileSystem(), 4000) as unknown as AbstractInputFileSystem,
-  extensions: ['.js', '.json'],
+  extensions: ['.ts', '.js', '.json'],
```
buschtoens: It's worth mentioning that the order is relevant here. The first match will be imported.

This means that _now_ , `.ts` files will take precedence. I think, we may want to put `.ts` last instead.

packages/ember-auto-import/ts/splitter.ts
```diff
@@ -17,7 +17,7 @@ const debug = makeDebug('ember-auto-import:splitter');
 const resolver = ResolverFactory.createResolver({
   // upstream types seem to be broken here
   fileSystem: new CachedInputFileSystem(new NodeJsInputFileSystem(), 4000) as unknown as AbstractInputFileSystem,
-  extensions: ['.js', '.json'],
+  extensions: ['.js', '.ts', '.json'],
```
buschtoens: The order in this array takes precedence over the order defined in [`webpack.resolve.extensions`](https://github.com/ef4/ember-auto-import/blob/89b72ea947e3fe35f1af91ba9df4d1a0bf83c797/packages/sample-typescript/ember-cli-build.js#L9-L11).

ef4:
Thanks, nice work!

jamescdavis:
Released in 1.6.0

Issue #234, opened 2019-07-17
`skipBabel` is missing from the documentation


Turbo87:
It seems that https://github.com/ef4/ember-auto-import/pull/225 introduced an option to opt-out of Babel transpilation for some dependencies, but it does not seem to be documented yet.

/cc @ef4 @pzuraq

danwenzel:
@ef4 @pzuraq  - ping ⬆️

ef4:
This is a good new contributor issue.

danwenzel:
@ef4 @pzuraq - Sure, I might take a crack at this if I get an extra moment.

Can you explain what the `semverRange` option is? https://github.com/ef4/ember-auto-import/pull/225/files#diff-2738f4eb29dd8800a17168c3223e173bR10

ef4:
Yeah, the idea is that you can say precisely what versions of the package should skip babel.

Given how NPM works, you can have many copies of the same library and they can be at different versions. This lets you get specific about which one is actually safe to use without going through babel.

danwenzel:
Looks like this was done via https://github.com/ef4/ember-auto-import/pull/252

Issue #251, opened 2019-10-24
`Uncaught ReferenceError: a is not defined` with mapbox-gl in production mode


kasunvp:
`a` in the error sometimes becomes `n`, I suppose due to dynamic naming at minification

Repo for reproduction: https://github.com/kasunvp/mapbox-gl-test (minimal app)

Error with `ember serve --prod` (can see the error even without the mapbox-gl access token)
Works with `ember serve` (have to add a mapbox-gl access token to load the map) 

The interesting fact is that we've been using ember-auto-import + mapbox-gl for some time now. Had a same kind of issue with a previous version, it was resolved and was working fine after using the ember-auto-import configuration `noParse: /(mapbox-gl)\.js$/` as was suggested somewhere. But not anymore with the latest versions.

Also, if the deprecated `ember-browserify` addon is used instead of ember-auto-import, it works fine too. So must be an issue in `ember-auto-import`/`webpack` combo.

Versions:
- Ember: v3.12.0
- ember-auto-import: v1.5.3
- mapbox-gl: v1.4.1

ef4:
Thanks for sharing a working reproduction, I can confirm I see the bug.

ef4:
Thanks for the update. Yes, skipBabel needs to get added to the readme, we have an open issue for that. I think it's a fine workaround to this problem. 

But I am still curious exactly what is breaking and I might dig in further. It's a little surprising that it breaks only in production, even after I remove uglify. That implies that either babel or webpack is doing something different in prod that results in invalid code.

kasunvp:
That maybe because mapbox is using a separate dev package for the dev build and a transpiled one in prod.

Also I included PR for the README change. Please let me know how it looks :)

PR #252, opened 2019-10-25
Solution for #251

kasunvp:
Solution for #251 

This PR adds couple of missing sections to the README that would fix
the above issue when used in an app or addon.

There is no code modification involved.


ef4:
Also fixes #234.

ef4:
Thank you!

jamescdavis:
Released in 1.6.0

PR #256, opened 2019-11-14
[DOC] Add dynamic import from addon instructions

### LINKED ISSUE #254: Unable to import dynamically in an addon
I'm attempting to auto import from within an addon component. I created [this small example](https://github.com/jrjohnson/reporoduction-auto-import-dynamic-addon) which has the code:
```
didInsertElement() {
    import("highcharts").then(() => {
      this.set('done', true);
    })
  }
```
 Running in the dummy app I get `TypeError: error resolving module specifier 'highcharts'` which I believe means that the import wasn't transpiled. I have 
```
babel: {
  plugins: [ require.resolve('ember-auto-import/babel-plugin') ]
}
```
in the addons `ember-cli-build.js` file, but I wonder if there is something different I need to do from an addon to modify the babel plugins.

jrjohnson:
Ember-cli-babel has the process for registering babel plugins worked out
for addons, adding a quick note here will save some pain in tracking
that information down.

Fixes #254

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,18 @@ Using ember-auto-import inside an addon is almost exactly the same as inside an 
       this._super.included.apply(this, arguments);
     }
     ```
+ - if your addon uses [Dynamic Import](#dynamic-import), it is [required](https://github.com/babel/ember-cli-babel#options) that you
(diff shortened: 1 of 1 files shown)


ef4:
Thanks, looks great.

PR #269, opened 2020-02-22
Explain the entry point in README

mehulkar:
@ef4 I'm not sure this is accurate, but I was trying to write a module to import and this is the behavior I observed. I haven't dug into how it works though, so let me know if I should update something.


ef4:
Thanks for this docs update, I think it's not quite the whole story and while we could expand coverage of how the main fields are chosen, it would need to be its own section -- this section is really just supposed to be an intro to the config.

PR #271, opened 2020-03-19
optional output fastboot files

houfeng0923:
@ef4 

there are many chunk files in my project, so auto-import-fastboot file size is too large .
if we never use fastboot , so ignore fastboot is better . 

related issue: https://github.com/ef4/ember-auto-import/issues/150

PR #272, opened 2020-03-19
Update node version to 12.x (latest LTS)

f1sherman:
Fixes the following CI error:
> error npm-package-arg@8.0.1: The engine "node" is incompatible with this module. Expected version ">=10". Got "8.17.0"

https://github.com/ef4/ember-auto-import/pull/271/checks?check_run_id=518716456

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -13,10 +13,10 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
-    - name: Use Node.js 8.x
+    - name: Use Node.js 12.x
       uses: actions/setup-node@v1
       with:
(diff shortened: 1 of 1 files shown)


f1sherman:
@ef4 sorry to bother you, is there anything else you'd like to see for this to be merged?

ef4:
Thanks!

f1sherman:
@Turbo87 this change only updated the version of node that is used in CI, it did not change anything in the library code

Turbo87:
sure, but if we no longer test on these old Node versions then how can you be sure that they still work?

f1sherman:
@Turbo87 that's a great point, I think a good enhancement would be to update CI to test all supported node versions.

Issue #275, opened 2020-04-08
CSP failure in consumeres of ember-auto-import 


sandstrom:
If a consuming app doesn't allow eval (via CSP) and the addon doesn't explicitly tell ember-auto-import to forbid eval, the compiled output will contain eval statements which will cause failure in the consuming app.

What's the performance cost? May I suggest that we drop eval altogether? Or at least always drop it when running within the context of an addon (vs an app).

ef4:
> The consuming app does not use ember-auto-import.

Even when the consuming app doesn't use ember-auto-import, the app is allowed to include configuration (including `forbidEval`) in order to control how its addons use ember-auto-import.

This got fixed in https://github.com/ef4/ember-auto-import/pull/213

sandstrom:
@ef4 That's good! 

But it's still a problem that when, for example, one of my addons switch to ember-auto-import my app breaks + it breaks in a fairly non-obvious way.

Is there a big performance difference when not using eval?

How about dropping eval when running within the context of an addon?

ef4:
Picking a different default when the app itself doesn't use ember-auto-import is a reasonable idea.

sandstrom:
@ef4 Yeah, that would be great I think!

sandstrom:
@ef4 I've made a stab at it here: https://github.com/ef4/ember-auto-import/issues/275

No idea if that's a good route to go down. Feel free to discard everything or make any changes you'd like to that PR. I've given you access to the branch, so edit at will.

Also not sure how one would check if the root package is using this addon or not! I guessed an empty options object may be an indicator, but I'm not sure about that.

jelhan:
> Picking a different default when the app itself doesn't use ember-auto-import is a reasonable idea.

Maybe the default could also depend on the presence of ember-cli-content-security-policy? There might be cases in which the app depends on the ember-cli-content-security-policy addon but it's disabled for local development. There might be even cases in which the app uses it but includes `'unsafe-eval'` in `script-src`. But both scenarios aren't the common ones.

adong:
Add the following config to `ember-cli-build.js` has worked for us.

```
let app = new EmberApp(defaults, {
    autoImport: {
      forbidEval: true
    },
```

PR #277, opened 2020-04-15
Fix CSP issue in apps using the addon indirectly

sandstrom:
If app doesn't use ember-auto-import directly we'll use 
an eval-safe source map generation strategy, since we
don't know if the app is using Content Security Policy (CSP) 
or not.

This may fix https://github.com/ef4/ember-auto-import/issues/275


ef4:
Thanks for submitting this and sorry for slow reply. 

A challenge here is that the `devtool: "source-map"` is both slower and interacts badly with ember-cli's own source mapping.  I would rather only do this when people are actually using CSP, and right now the reliable solution is to make apps configure it.

I will also comment on #275 about the general strategy here.

PR #279, opened 2020-04-21
fix: support multiple projects in one process

### LINKED ISSUE #278: `AutoImport.lookup(appOrAddon)` doesn't support multi-project builds
`AutoImport.lookup(appOrAddon)` doesn't support building multiple projects in the same process, registering on `global` assumes that you'll only have one [`Project`](https://ember-cli.com/api/classes/Project.html) instance per Node process.

https://github.com/ef4/ember-auto-import/blob/74dcdf289bdb03542f345a1ad8f660558fdc61be/packages/ember-auto-import/ts/auto-import.ts#L22-L28

In my case I am pre-rendering a little dummy app via `ember-cli-fastboot` and inject the output HTML into the host app's `index.html` at build time.

The dummy app fails to render, if it tries to import anything via `ember-auto-import` as the imports are registered at the host app instead.

buschtoens:
Fixes #278.

I'm not too happy with how this turned out. If you have a better idea, I would highly appreciate hearing it!

The tests also seem to be failing locally, but it appears to be an unrelated issue with my setup.

### DIFF
--- a/packages/ember-auto-import/ts/auto-import.ts
+++ b/packages/ember-auto-import/ts/auto-import.ts
@@ -8,7 +8,8 @@ import Append from './broccoli-append';
 import { Tree } from 'broccoli-plugin';
 
 const debugTree = buildDebugCallback('ember-auto-import');
-const protocol = '__ember_auto_import_protocol_v1__';
+const protocolV1 = '__ember_auto_import_protocol_v1__';
+const protocolV2 = '__ember_auto_import_protocol_v2__';
 
 export default class AutoImport {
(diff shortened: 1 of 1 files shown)


buschtoens:
There seems to be some build flakiness.

buschtoens:
I've rebased the PR onto the latest `master` and removed all the unrelated extra fixes which are already included in `master`.

ef4:
Thanks for taking a stab at this, I'm going to close in favor of a bigger refactor in #325 that can also ensure the newest copy becomes the leader.

PR #281, opened 2020-04-24
add [id] to the chunkname

### LINKED ISSUE #280: Webpack produced output is not deterministic
The issue is described here https://github.com/ember-cli/ember-cli/issues/9187

I could narrow it down to `ember-auto-import` and its webpack configuration. The problem is following:
- the webpack is configured to have these chunk names:

```
      output: {
        filename: `chunk.[chunkhash].js`,
        chunkFilename: `chunk.[chunkhash].js`,
      },
```
- so it substitutes the chunk hash which could be different and it really depends on the a source code sits in, because source files have the absolute paths in them

```
if (typeof document !== 'undefined') {
  __webpack_public_path__ = (function(){
    var scripts = document.querySelectorAll('script');
    return scripts[scripts.length - 1].src.replace(/\/[^/]*$/, '/');
  })();
}

module.exports = (function(){
  var d = _eai_d;
  var r = _eai_r;
  window.emberAutoImportDynamic = function(specifier) {
    return r('_eai_dyn_' + specifier);
  };
    d('@glimmer/tracking', [], function() { return require('D:\\Projects\\_something\\test-ember-asset-build-order\\node_modules\\@glimmer\\tracking\\dist\\modules\\es2017\\index.js'); });
    d('macro-decorators', [], function() { return require('D:\\Projects\\_something\\test-ember-asset-build-order\\node_modules\\macro-decorators\\dist\\index.js'); });
})();  
```

- then the chunks are taken, I assume, simply in the alphabetical order

I've created a small temporary workaround for the problem for our project which implies simply adding of the `[id]` variable in the filename. I'll create a PR here too.

The question to maintainers: is that required to have it configured this way? Why don't we simply use [webpack's default](https://v4.webpack.js.org/configuration/output/#outputchunkfilename) `[id].js`?

stukalin:
Fixes #280

### DIFF
--- a/packages/ember-auto-import/ts/webpack.ts
+++ b/packages/ember-auto-import/ts/webpack.ts
@@ -95,8 +95,8 @@ export default class WebpackBundler implements BundlerHook {
       },
       output: {
         path: this.outputDir,
-        filename: `chunk.[chunkhash].js`,
-        chunkFilename: `chunk.[chunkhash].js`,
+        filename: `chunk.[id].[chunkhash].js`,
+        chunkFilename: `chunk.[id].[chunkhash].js`,
         libraryTarget: 'var',
         library: '__ember_auto_import__'
       },


ef4:
Thanks, this seems fine. I'll see about why CI is failing for this PR, because I doubt it's really due to this change.

ef4:
(that force push is a rebase to master)

jamescdavis:
Released in 1.6.0

PR #289, opened 2020-06-04
Update minimum version of @babel/preset-env to 7.10.2.

rwjblue:
`@babel/preset-env` leverages two lists of plugins to determine what to _actually_ do for any given set of targets. The first list comes from the gathered browser compat data and lives in `@babel/compat-data`, and the second list is the actually known set of plugins that `@babel/preset-env` itself depends on. Additionally, `@babel/preset-env`'s dependency on `@babel/compat-data` is a `^` dependency very intentionally allowing SemVer drift (so you can get the latest set of known support for browsers).

Unfortunately, prior to `@babel/preset-env@7.8.7` it would take the list of plugins directly from `@babel/compat-data` and attempt to compile them. Due to the fact that `@babel/compat-data` can drift, you can end up in a situation where the list of plugins that are attempted include plugins that the version of `@babel/preset-env` that `ember-auto-import` is using has **no clue** about.

This presents with an error message like the following:

```
ERROR in ./node_modules/@glimmer/tracking/dist/modules/es2017/index.js
Module build failed (from ./node_modules/ember-auto-import/node_modules/babel-loader/lib/index.js):
Error: [BABEL] /Users/rjackson/src/my-app-here/node_modules/@glimmer/tracking/dist/modules/es2017/index.js: Could not find plugin "proposal-numeric-separator". Ensure there is an entry in ./available-plugins.js for it. (While processing: "/Users/rjackson/src/my-app-here/node_modules/ember-auto-import/node_modules/@babel/preset-env/lib/index.js")
    at getPlugin (/Users/rjackson/src/my-app-here/node_modules/ember-auto-import/node_modules/@babel/preset-env/lib/index.js:68:11)
```

The underlying issue was fixed in `@babel/preset-env` in https://github.com/babel/babel/pull/11201 (which was included in `@babel/preset-env@7.8.7`).

Paired with @hjdivad to track this down.

### DIFF
--- a/packages/ember-auto-import/package.json
+++ b/packages/ember-auto-import/package.json
@@ -28,7 +28,7 @@
   },
   "dependencies": {
     "@babel/core": "^7.1.6",
-    "@babel/preset-env": "^7.0.0",
+    "@babel/preset-env": "^7.10.2",
     "@babel/traverse": "^7.1.6",
     "@babel/types": "^7.1.6",
     "@embroider/core": "^0.4.3",
--- a/yarn.lock
+++ b/yarn.lock
@@ -16,12 +16,28 @@
   dependencies:
     "@babel/highlight" "^7.0.0"
 
+"@babel/code-frame@^7.10.1":
+  version "7.10.1"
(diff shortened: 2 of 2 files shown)


jamescdavis:
Released in 1.6.0

PR #293, opened 2020-06-15
support dynamic import from template urls

jelhan:
This adds support for dynamic `import()` using absolute URLs as described in v2 addon format RFC:

> To be more precise, the only supported syntax inside `import()` is:
> - ...
> - or a template string that begins with a static prefix which falls into one of these allowed cases:
>   - the static prefix matches /(\w+:)?\/\//
>     - this detects absolute URLs, both ones with protocols (starting with http://) and ones that are protocol-relative (starting with //).
>     - in this case, Embroider will leave your import() alone so you can access the browser's implementation of import(). You are responsible for the contents of the URL, it is beyond the scope of the Embroider build.
>   - ...

It also improves the error reporting if user tries to use dynamic `import()` syntax, which is allowed by v2 addon format but not yet supported by ember-auto-import.

It includes a test that usage of dynamic `import()` with absolute URLs that include a dynamic part is supported. But it does _not_ add tests that build throws on invalid usage of dynamic `import()`. I hope that's okay as there doesn't seem to be any such test yet at all.


packages/ember-auto-import/ts/analyzer.ts
```diff
@@ -162,11 +162,68 @@ export default class Analyzer extends Plugin {
...
+            ) {
+              // In this case, Embroider will leave the import() alone. The
+              // browser's implementation of import() is used. The contents of
+              // the URL is beyond the scope of the Embroider build.
+              return;
+            } else if (
+              // The static prefix is a NPM package name or a relative path.
+              (
+                // This matches a namespaced NPM package name.
+                prefix.startsWith('@') && (prefix.match(/\//) || []).length >= 2
```
jamescdavis, line 201: I don't think it does. The length of result of `match` will never be > 1 because it doesn't capture anything. How about this instead:
```suggestion
                /^@[^/]+\/[^/]+$/.test(prefix)
```

packages/ember-auto-import/ts/analyzer.ts
```diff
@@ -162,11 +162,68 @@ export default class Analyzer extends Plugin {
...
+              ) ||
+              (
+                // This matches a NPM package name without namespace and a
+                // relative path.
+                !prefix.startsWith('@') && prefix.includes('/')
+              )
+            ) {
+              throw new Error(
+                'ember-auto-import does not support dynamic import() with a ' +
+                'template literal that references a NPM package or a relative' +
```
jamescdavis, line 211:
```suggestion
                'template literal that references an NPM package or a relative' +
```

packages/ember-auto-import/ts/analyzer.ts
```diff
@@ -162,11 +162,68 @@ export default class Analyzer extends Plugin {
...
+                "path yet even so it's allowed by v2 addon format."
```
jamescdavis, line 212:
```suggestion
                "path yet even though it's allowed by v2 addon format."
```

packages/ember-auto-import/ts/analyzer.ts
```diff
@@ -162,11 +162,68 @@ export default class Analyzer extends Plugin {
...
+              );
+            } else {
+              // Only template literals that match one of the cases tested
+              // before are supported.
+              throw new Error(
+                'ember-auto-import only supports dynamic import() that are ' +
+                'included in the supported subset of dynamic import syntax ' +
+                'of v2 addon format. A template literal argument must begin ' +
+                'with a static prefix which either identifies it as (1) an ' +
+                'absoulte URL, (2) a npm package or (3) a relative path.'
```
jamescdavis, line 222: (2) and (3) are not supported, right? So we shouldn't list them here?

packages/ember-auto-import/ts/analyzer.ts
```diff
@@ -162,11 +162,68 @@ export default class Analyzer extends Plugin {
...
             throw new Error(
-              'ember-auto-import only supports dynamic import() with a string literal argument.'
+              'ember-auto-import only supports dynamic import() that are ' +
+              'included in the supported subset of dynamic import syntax ' +
+              'of v2 addon format. Only string or template literal arguments ' +
+              `are allowed by that specification but ${argument.type} is used.`
             );
           }
+
+          // Only template strings that begins with a static prefix which
```
jamescdavis, line 178:
```suggestion
          // Only template strings that begin with a static prefix which
```

packages/ember-auto-import/ts/analyzer.ts
```diff
@@ -162,11 +162,68 @@ export default class Analyzer extends Plugin {
...
+              // the URL is beyond the scope of the Embroider build.
+              return;
+            } else if (
+              // The static prefix is a NPM package name or a relative path.
+              (
+                // This matches a namespaced NPM package name.
+                prefix.startsWith('@') && (prefix.match(/\//) || []).length >= 2
+              ) ||
+              (
+                // This matches a NPM package name without namespace and a
```
jamescdavis, line 204:
```suggestion
                // This matches an NPM package name without namespace and a
```

packages/ember-auto-import/ts/analyzer.ts
```diff
@@ -162,11 +162,68 @@ export default class Analyzer extends Plugin {
...
+                // relative path.
+                !prefix.startsWith('@') && prefix.includes('/')
```
jamescdavis, line 206: This wouldn't match an NPM package name without a namespace because it wouldn't have a `/` in it.

packages/ember-auto-import/ts/analyzer.ts
jelhan: You are right. It was missing the global modifier:

```js
let prefix = '@foo/ab/';
prefix.match(/\/).length // 1
prefix.match(/\//g).length // 2
```

```suggestion
                prefix.startsWith('@') && (prefix.match(/\//g) || []).length >= 2
```

I decided for this implementation cause it's following the RFC closely:

> - the static prefix starts with @ and contains at least two / 
>   - this detects a namespaced NPM package name. If the package can't be resolved or it contains zero files that match your pattern, we emit a static build error.

I think your regular expression is not testing if the string contains a closing slash at the end. The prefix must not only be a valid NPM package name. It must also include a slash that marks the end of the package name. Otherwise the pattern would allow invalid patterns like this:

```js
let invalid = `@foo/bar${dynamic}`;
```

It should also allow additional characters as something like this is fully valid:

```js
let valid = `@foo/bar/baz/${dynamic}`;
```

Additionally it seems to miss some escaping of slashes. I think a working solution would look like this:

```suggestion
                /^@[^\/]+\/[^\/]+\//.test(prefix)
```

Not sure which one to prefer.

packages/ember-auto-import/ts/analyzer.ts
jelhan: I fear the comment is misleading. We try to match a NPM package including a closed slash:

```js
let valid = `foo/${dynamic}`;
let invalid = `foo${dynamic}`;
```

packages/ember-auto-import/ts/analyzer.ts
jelhan: I liked to explicitly mention the supported subset of v2 addon format in that error message. But I see that it might be frustrating for consumers to than refactor to (2) or (3) just to notice that it's not supported yet. I tried to improve that error message in 74c73bd589c3a25dbfea797bd7d5ea5c27c38b0f to mention both: Syntax allowed by v2 addon format and syntax supported by ember-auto-import so far.

jelhan:
While going through the review comments I noticed that the edge case of a template literal without any expression wasn't handled well yet. Adding support for that one would require some additional complexity. I think it's okay to just ask the user to use a string instead. I added such an error message in c6a6de2605bff1c01021f9cbef467b5d59f4e126.

jelhan:
Rebased to fix the merge conflict in `yarn.lock`. Please let me know if there is anything l can support you with to get this one merged.

packages/ember-auto-import/ts/analyzer.ts
jelhan: Applied the fix to my preferred solution so that this one could be merged without any additional changes. Let me know if you still prefer the regular expression one. Open to refactor in that case.

jelhan:
@jamescdavis Any chance you can give this another review?

jelhan:
@ef4 Thanks for merging latest master in. Is there anything I can do to help getting it merged?

ef4:
Thanks for working on this, and apologies that I haven’t had time sooner to give clear direction. 

This PR doesn’t solve the problem for anybody who is also using ember-auto-import’s Babel plugin for dynamic imports, because that will still rewrite them unconditionally to call our own runtime implementation.

Also it’s hard to test the error cases using an end-to-end test like this. They can use node unit tests. 

But also, I am trying to implement the full spec for template dynamic import, not just URLs, and that implies that the analyzer needs to share info about template literals and leave the deciding up to the splitter. 

I have a branch building off yours, see template-string-import.

Issue #284, opened 2020-05-08
Imported module is bundled into vendor.js instead of test-support.js


simonihmig:
It seem under a certain condition it is possible that an imported module is put into `vendor.js` in instead of `test-support.js`, making it get shipped to users. 😟

This happens when:
* the addon has an `treeForAddonTestSupport` hook that removes the `/test-support` module namespace, by not calling `_super`. Which a lot of addons that are focused on testing only do, e.g.  https://github.com/emberjs/ember-qunit/blob/master/index.js#L78-L81
* in the `/addon-test-support` tree there is an import statement for a npm package handled by `ember-auto-import`

The concrete case where I witnessed this is for `ember-cli-yadda`, where the above conditions are fulfilled:
* https://github.com/albertjan/ember-cli-yadda/blob/master/index.js#L48
* https://github.com/albertjan/ember-cli-yadda/blob/master/addon-test-support/index.js

When I remove the overridden `treeForAddonTestSupport`, the `yadda` library is added to `test-support.js`. And when I remove the [import](https://github.com/albertjan/ember-cli-yadda/blob/master/addon-test-support/index.js) then obviously `yadda` is not imported ata ll (ensuring this was the offending `import` that caused the incorrect bundling)

No idea if this is easy or hard to fix? For the current case I believe we can workaround this pretty well. But still think this looks like a bug here.

ef4:
Thanks for reporting Simon. The code that controls this is in
bundle-config.ts. There’s a regexp that looks for test-support. That
strategy probably fails in this case.

It would be better if we could tell which tree we were preprocessing, but I
don’t know if preprocessor are given that info. One place to poke around is
if our toTree method has access to any additional info we could capture.

simonihmig:
I poked around a bit in node debugger, but there seems to be nothing reliable we can use here. At some early point in the call stack the addon is called with `addon.treeFor(type)` (with `type` being `"addon-test-support"` here), but that info is lost from then on...

Adding that kind of meta data to the trees would be some pretty invasive change I think, requiring updates to `ember-cli` itself. Also probably a controversial change...

What we could do quite easily I think is to add some meta data to the options passed to `preprocessJs()` here: https://github.com/albertjan/ember-cli-yadda/blob/master/index.js#L64. Something like:

```js
return this.preprocessJs(namespacedTree, '/', this.name, {
  registry: this.registry,
  autoImport: {
    bundleType: 'test'
  }
});
```

[...]

PR #294, opened 2020-06-16
Update regex to work with scoped npm packages

paddyobrien:
Towards https://github.com/ef4/ember-auto-import/issues/284

This regex is used to detect if an import is referencing the tests or addon-test-support folders.

As it stands it will match `my-package/test-support` but not `@scoped/my-package/test-support` which seems like an oversight.

I honestly don't fully understand the potential downstream consequences of this or how to test it, but it seems like a fairly uncontroversial fix. I took a look at the tests and they're a little hard to follow, (which is understandable given the nature of this library), happy to add one if you can point me in the right direction.

### DIFF
--- a/packages/ember-auto-import/ts/bundle-config.ts
+++ b/packages/ember-auto-import/ts/bundle-config.ts
@@ -4,7 +4,7 @@
 */
 
 import { dirname } from 'path';
-const testsPattern = new RegExp(`^/?[^/]+/(tests|test-support)/`);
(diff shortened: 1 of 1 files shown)


ef4:
I think this is probably OK but I've held off on merging it because I haven't had a chance to add test coverage, and it does deserve test coverage.

Also, this problem may stop mattering due to https://github.com/ef4/ember-auto-import/pull/329, but not until ember-cli releases its corresponding change https://github.com/ember-cli/ember-cli/pull/9411

PR #301, opened 2020-07-24
Update CONTRIBUTING.md to make sure people cd into the right place.

kiwiupover:


ef4:
I think technically this would be `cd ember-auto-import/packages/ember-auto-import`, since the original instructions assumed you just can `git clone`.

PR #304, opened 2020-08-14
Fix merging addon webpack configs

bendemboski:
Like we do [here](https://github.com/ef4/ember-auto-import/blob/ab4cc5614bfb7b56e0649de917b3de8bc073461f/packages/ember-auto-import/ts/webpack.ts#L229-L242), we need to concatenate arrays here or addons and apps will overwrite each other's array configurations.

I'm not sure how best to write a test for this change, and would love some advice.


ef4:
Thanks

PR #308, opened 2020-09-11
Update @embroider/core dependency

simonihmig:
Should hopefully fix https://github.com/embroider-build/embroider/issues/509

### DIFF
--- a/packages/ember-auto-import/tsconfig.json
+++ b/packages/ember-auto-import/tsconfig.json
@@ -12,6 +12,7 @@
     "noUnusedParameters": true,
     "sourceMap": true,
     "experimentalDecorators": true,
-    "strict": true
+    "strict": true,
+    "skipLibCheck": true
   }
 }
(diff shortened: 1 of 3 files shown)


packages/ember-auto-import/tsconfig.json
```diff
@@ -12,6 +12,7 @@
     "noUnusedParameters": true,
     "sourceMap": true,
     "experimentalDecorators": true,
-    "strict": true
+    "strict": true,
+    "skipLibCheck": true
```
simonihmig, line 16: I had to add this to fix `tsc` choking on the new embroider supplied types:

```
../../node_modules/@embroider/core/src/wait-for-trees.d.ts:6:5 - error TS2416: Property 'build' in type 'WaitForTrees<NamedTrees>' is not assignable to the same property in base type 'Plugin'.
  Type '(detail: { changedNodes: boolean[]; } | undefined) => Promise<void>' is not assignable to type '() => void | Promise<void>'.
```

🤔

ef4:
Thanks! I think `skipLibCheck` in this context is OK.

simonihmig:
@ef4 could we please get a release of this? The issue linked above is still present without a release unfortunately...

PR #312, opened 2020-09-29
upgrade ember-cli-babel to v7

nlfurniss:
Also needed to move Node support to 10, the oldest still maintained version.


ef4:
This seems good, but we need passing tests.

nlfurniss:
@ef4 fixed the issue that caused the original failures, but it looks like the canary one wasn't able to start. Mind re-running the checks?

ef4:
Thank you!

PR #319, opened 2020-11-18
Allow semver matching versions of libraries to pass the safe version check

omarestrella:
This is re-opening #236, as the original repository seems to have disappeared and Github does not allow editing the repository for pull requests.

For #202, just to get the ball rolling.

Reads from the package.json in order to perform a semver check whenever you have a shared sub-dependency with one of your own dependencies.


omarestrella:
Have been running this change internally for >6 months now, but would like to formalize it via a test. I seem to be having trouble with running the tests locally using the included script. Is there a way I can test this individually in one of the included samples (by modifying them or creating a new one)?

ef4:
Thanks for taking a stab at this, and apologies for not giving feedback sooner. Partly because of changes since this pr was written, there are a couple things missing:

 - this won't detect conflicts between a copy of the package that's imported via string literal and one that's imported via template string literal, because those go down different paths. In practice, those go down such different paths that we *could* just allow the two different copies, but I think that would be confusing, so we should probably still assert.
 - we don't need to use `pkg-up` here because [target here](https://github.com/ef4/ember-auto-import/pull/319/files#diff-8ff68626dfb32cfc5db52d578b84fe00d7a5d0e30632da3028b9ce05cf88c390R99) already contains the name and location of the package that owns the entrypoint file.

I'm going to close this because I let it get so stale, but please feel free to open a fresh pr if you're still available to work on this.

PR #324, opened 2020-12-03
starting v2 addon support

ef4:


ef4:
(This depends on unreleased changes in embroider.)

ef4:
This is an old PR that was superseded. As of 2.0 we support v2 addons.

Issue #321, opened 2020-11-30
Webpack 5 support?


patocallaghan:
Webpack 5 was released ~1.5 months back ([release post](https://webpack.js.org/blog/2020-10-10-webpack-5-release/)) with a lot of new features. Have you had any thoughts on upgrading the webpack dependency to v5?

I'm quite curious about the new [persistent caching feature](https://webpack.js.org/blog/2020-10-10-webpack-5-release/#persistent-caching) and I attempted a spike over [here](https://github.com/ef4/ember-auto-import/compare/master...patocallaghan:patoc/webpack-5) (i.e. did the bare minimum to make it compile) and try it out in our quite large app through `yarn link`. I configured it to use the persistent caching settings in the `autoImport.webpack` config in ember-cli-build.js, and I see a consistent 10-15 second speed boost on warm build startups as the `Bundler` step is reduced to just a few seconds with a primed cache.

Here's an extremely [simplified app](https://github.com/patocallaghan/deps-app) with [a few deps](https://github.com/patocallaghan/deps-app/blob/d2b75c399e445a4ef74180e1e5baa660487f01dc/app/controllers/application.js#L2-L20) and the [persistent caching config](https://github.com/patocallaghan/deps-app/blob/d2b75c399e445a4ef74180e1e5baa660487f01dc/ember-cli-build.js#L8-L14). As you can see there's a 5 second boost to the warm build startup time, and this is a tiny app.

### Webpack 4
![image](https://user-images.githubusercontent.com/685034/100671036-b4988d80-3357-11eb-8bad-5b154d016714.png)

### Webpack 5 + persistent cache
![image](https://user-images.githubusercontent.com/685034/100671082-c712c700-3357-11eb-970b-18a5d4b7b34b.png)

I guess there would be some things to consider:
1. There doesn't appear to be any breaking changes with regards the [configuration options ember-auto-import uses](https://webpack.js.org/migrate/5/#update-outdated-options) in the [internal webpack config](https://github.com/ef4/ember-auto-import/blob/7610021b3c592bcaa5a649a1d6fa799826f857d3/packages/ember-auto-import/ts/webpack.ts#L90-L121)
2. While webpack is an internal dependency, consuming apps may configure webpack themselves using the `autoImport.webpack` config so this would be a breaking change for them
[...]

PR #327, opened 2020-12-07
[chore] bump webpack to v5

gabrielcsapo:
*Motivation*

fixes https://github.com/ef4/ember-auto-import/issues/321

webpack@5 started to parse window.require, so using _non_webpack_require will tell webpack to ignore this during the build step and execute native require when executed.


patocallaghan:
Nice @gabrielcsapo 🙌 It looks like CI is failing though 🤔

ef4:
Looks like fastboot is broken due to some changes in webpack's publicPath handling.

gabrielcsapo:
Looking into fixing CI and fastboot now.

gabrielcsapo:
I found a fix for the publicPath issue and am chasing down the remaining errors now

gabrielcsapo:
Updated with the latest and double checking if this fixed the failing tests.

gabrielcsapo:
@ef4 @thoov it passes!

ef4:
Nice work, thanks.

gabrielcsapo:
Any update on merging this @ef4 ?

ef4:
I was considering updating `@embroider/webpack` simultaneously, so there's one less difference between classic builds with ember-auto-import and embroider builds. But I can't actually think of a hard reason to wait for that, so this can merge and release.

This will be a semver major. But given the only-one-can-win nature of emer-auto-import, there's no way to release it without risking breakage. If an addon and the app ask for different major versions, we can't satisfy both, and instead we always let the newer one take charge. 

The breaking changes between webpack 4 and 5 don't seem very big, at least in terms of the API that typical ember apps configure. So hopefully this goes smoothly.

yads:
@ef4 friendly nudge on getting this merged in. Is there something we can do to help?

ef4:
Yes, I would prefer to not break apps that have custom webpack config just because one of their addons upgrades ember-auto-import. I think we should improve the leader election code to help.

 - in apps that don't directly depend on ember-auto-import, things can work as they do now.
 - in apps that don't pass any custom webpack config to ember-auto-import (or that only pass things we know are still OK in webpack 5), things can also continue to work as they do now
 - otherwise, we should restrict the leader election so that only copies of ember-auto-import that are semver-compatible with the app's copy can win
 - this may result in an addon getting an older major release of ember-auto-import than it requires, which should result in a build error.

This implies that addons will have a minimum-supported ember-auto-import major version, and if they want to increase it they should do their own semver major releases in order to not break apps.

ef4:
Getting ready to merge this because we now have [a 2.0 rollout plan](https://github.com/ef4/ember-auto-import/pull/379).

The remaining test failures seem to be that we're loading test-only dependencies into fastboot.

gabrielcsapo:
@ef4 looking to fix these issues now.

gabrielcsapo:
@ef4 failures seem to be caused from moving around the sample app tests and not using the right version of fast boot.

gabrielcsapo:
@ef4 after the tests were refactored something has changed so that now these are never passing as a result with the webpack@5 changes...

ef4:
Thanks again @gabrielcsapo. There was one real bug caused by a change in webpack's stats format, but I was able to find it.

gabrielcsapo:
You sir are a hero 🙌

PR #328, opened 2020-12-08
Migrate from resolve.sync and pkgUp to resolve-package-path

rwjblue:
`resolve-package-path` solves the same basic purpose of `resolve.sync` and `pkgUp` but introduces caching to avoid repeated lookups of the same package name. Migrating to `resolve-package-path` will reduce overall I/O cost added by ember-auto-import since a number of other common pieces in the addon ecosystem use `resolve-package-path` (e.g.  `ember-cli-version-checker`) and these tools can now share the same cache (avoiding wasted I/O).

See [stefanpenner/resolve-package-path](https://github.com/stefanpenner/resolve-package-path) for details.


packages/ember-auto-import/ts/package.ts
```diff
@@ -159,9 +159,13 @@ export default class Package {
 
   isEmberAddonDependency(name: string): boolean {
     if (!this.isAddonCache.has(name)) {
-      let packageJSON = require(resolve.sync(`${name}/package.json`, {
-        basedir: this.root
-      }));
+      let packagePath = resolvePackagePath(`${name}/package.json`, this.root);
+
+      if (packagePath === null) {
+        throw new Error(`${ this.name } tried to import "${name}" but the package was not resolvable from ${this.root}`);
```
rwjblue, line 165: This error is added mostly to satisfy typing constraints (since `resolvePackagePath` may return `null` if the package is not found). This does **not** introduce additional cases where this would happen though, `resolve.sync` just threw an error internally if it couldn't resolve the package.

PR #329, opened 2020-12-09
Infer bundle type (app vs test) by original tree if known.

rwjblue:
Prior to these changes, we would decide what bundle a given file should end up in based on the modules relativePath that imported it. This worked generally for application level tests that import external packages, as well as for many addon-test-support imports.

Unfortunately, this relative path analysis falls down in addons that intentionally customize their `treeForAddonTestSupport` so that the `addon-test-support/` folder contents are available **without** the `addon-name/test-support` suffix. This is done by a number of testing only addons, but the largest offenders are `ember-qunit` and `@ember/test-helpers`.

This introduces a new (possibly undefined) value that can be passed into `preprocessJs` to specify the `treeType` being transpiled. ember-cli will update to pass this value by default, but any addons doing the sort of overrides mentioned above are _likely_ already doing a manual override and can easily pass in the treeType in their overrides.

Fixes https://github.com/ef4/ember-auto-import/issues/284

### DIFF
--- a/packages/ember-auto-import/ts/analyzer.ts
+++ b/packages/ember-auto-import/ts/analyzer.ts
@@ -6,7 +6,7 @@ import FSTree from 'fs-tree-diff';
 import makeDebug from 'debug';
 import { join, extname } from 'path';
 import { isEqual, flatten } from 'lodash';
-import Package from './package';
+import type Package from './package';
 import symlinkOrCopy from 'symlink-or-copy';
 import { TransformOptions } from '@babel/core';
 import { File } from '@babel/types';
@@ -18,7 +18,8 @@ makeDebug.formatters.m = (modules: Import[]) => {
       specifier: m.specifier,
       path: m.path,
       isDynamic: m.isDynamic,
-      package: m.package.name
+      package: m.package.name,
+      treeType: m.treeType
     })),
     null,
     2
@@ -27,11 +28,14 @@ makeDebug.formatters.m = (modules: Import[]) => {
 
 const debug = makeDebug('ember-auto-import:analyzer');
 
+export type TreeType = 'app' | 'addon' | 'addon-test-support' | 'test';
+
 export interface Import {
   path: string;
   package: Package;
   specifier: string;
   isDynamic: boolean;
+  treeType: TreeType | undefined;
 }
 
 /*
@@ -45,7 +49,7 @@ export default class Analyzer extends Plugin {
 
   private parse: undefined | ((source: string) => File);
 
-  constructor(inputTree: Node, private pack: Package) {
+  constructor(inputTree: Node, private pack: Package, private treeType?: TreeType) {
     super([inputTree], {
       annotation: 'ember-auto-import-analyzer',
       persistentOutput: true
@@ -172,7 +176,8 @@ export default class Analyzer extends Plugin {
             isDynamic: true,
             specifier: argument.value,
             path: relativePath,
-            package: this.pack
+            package: this.pack,
+            treeType: this.treeType,
           });
         }
       },
@@ -181,7 +186,8 @@ export default class Analyzer extends Plugin {
           isDynamic: false,
           specifier: path.node.source.value,
           path: relativePath,
-          package: this.pack
+          package: this.pack,
+          treeType: this.treeType,
         });
       },
       ExportNamedDeclaration: (path) => {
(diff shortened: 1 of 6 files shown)


ef4:
Thanks for working on this. Would this allow an addon to have some files of each kind? I was picturing an API more like:

```js
// my-addon/index.js
module.exports = {
  name: 'my-addon',
  options: {
    autoImport:{
      testsPattern: /\/custom-test-support\//
    }
  }
};
```

`testsPattern` would override https://github.com/ef4/ember-auto-import/blob/b558c7c0c53912e03b3f7c7df0e998154cf0cf04/packages/ember-auto-import/ts/bundle-config.ts#L8, but only for this package.

We already track which package goes with each tree that's being analyzed.

rwjblue:
> Would this allow an addon to have some files of each kind?

Ya, if the import is in `addon` tree those imports would go to `assets/vendor.js` and if the import is in `addon-test-support` the imports would end up in the `assets/test-support.js` bundle.

> I was picturing an API more like:
> 
> ```js
> // my-addon/index.js
> module.exports = {
>   name: 'my-addon',
>   options: {
>     autoImport:{
>       testsPattern: /\/custom-test-support\//
>     }
>   }
> };
> ```

Ya, I can do this also as a way to override completely, but I generally would prefer to _default_ to not needing the custom test pattern (and instead infer if a given import should be vendor vs test-support based on the source tree).

ef4:
Ok, yeah, as long as you're fine requiring the ember-cli change we can stick with that plan and don't need to also introduce the option to override the pattern.

rwjblue:
Ya, I have a branch with the ember-cli changes required all queued up. However, it is also possible for individual addons to **guarantee** they are compatible without requiring the ember-cli change. 

For example [this code in ember-qunit](https://github.com/emberjs/ember-qunit/blob/d15a4ab1c6e9091ed06415d469c5e8894355ceeb/index.js#L79-L82):

```js
  treeForAddonTestSupport(tree) {
    // intentionally not calling _super here
    // so that can have our `import`'s be
    // import { ... } from 'ember-qunit';

    const Funnel = require('broccoli-funnel');

    let scopedInputTree = new Funnel(tree, { destDir: 'ember-qunit' });

    return this.preprocessJs(scopedInputTree, '/', this.name, {
      annotation: `ember-qunit - treeForAddonTestSupport`,
      registry: this.registry,
    });
  },
```

Will turn into:

```js
  treeForAddonTestSupport(tree) {
    // intentionally not calling _super here
    // so that can have our `import`'s be
    // import { ... } from 'ember-qunit';

    const Funnel = require('broccoli-funnel');

    let scopedInputTree = new Funnel(tree, { destDir: 'ember-qunit' });

    return this.preprocessJs(scopedInputTree, '/', this.name, {
      annotation: `ember-qunit - treeForAddonTestSupport`,
      registry: this.registry,
      treeType: 'addon-test-support'
    });
  },
```

rwjblue:
@scalvert helped me track down the cause of the `tsc` failure, so I think this should pass CI now...

scalvert:
I think it may be nice to support both `testsPattern` as a "get out of jail free card", while primarily using `treeType`. That would allow addons that control the invocation of the preprocessor to use the latter, while giving us an "out" if things go awry.

Nice work on this, @rwjblue.

packages/ember-auto-import/ts/analyzer.ts
```diff
@@ -6,7 +6,7 @@ import FSTree from 'fs-tree-diff';
 import makeDebug from 'debug';
 import { join, extname } from 'path';
 import { isEqual, flatten } from 'lodash';
-import Package from './package';
+import type Package from './package';
```
ef4, line 9: I'm really glad typescript added this feature.

packages/ember-auto-import/ts/auto-import.ts
```diff
@@ -16,7 +17,7 @@ const debugTree = buildDebugCallback('ember-auto-import');
 // what you're doing.
 export interface AutoImportSharedAPI {
   isPrimary(addonInstance: AddonInstance): boolean;
-  analyze(tree: Node, addon: AddonInstance): Node;
+  analyze(tree: Node, addon: AddonInstance, treeType?: TreeType): Node;
```
ef4, line 20: This area deserves extra scrutiny because different versions of ember-auto-import will call each other's copy of `analyze`. But this looks safe to me -- older copies will not pass the optional argument, newer copies will pass the argument and it will be ignored.

ef4 (approved):

PR #330, opened 2020-12-11
Ensure `bundleForTreeType` returns the proper values.

rwjblue:
When introducing `bundleForTreeType` (in #329) I accidentally emitted an invalid bundle type (`test` vs the correct `tests`). This updates to the correct value, extracts a shared type (to confirm things stay in sync), and adds type assertions to ensure that all cases are handled in both `bundleEntrybundleEntrypoint` and `bundleForPath`.

### DIFF
--- a/packages/ember-auto-import/ts/bundle-config.ts
+++ b/packages/ember-auto-import/ts/bundle-config.ts
@@ -59,6 +59,9 @@ export default class BundleConfig {
     switch (treeType) {
       case 'app':
       case 'addon':
+      case 'addon-templates':
+      case 'styles':
+      case 'templates':
         return 'app';
 
       case 'addon-test-support':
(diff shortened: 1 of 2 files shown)


packages/ember-auto-import/ts/bundle-config.ts
```diff
@@ -9,61 +9,70 @@ const testsPattern = new RegExp(`^/?[^/]+/(tests|test-support)/`);
...
+  bundleForTreeType(treeType: TreeType): BundleName {
     switch (treeType) {
       case 'app':
       case 'addon':
         return 'app';
 
       case 'addon-test-support':
       case 'test':
-        return 'test';
+        return 'tests';
```
rwjblue, line 66: This is the real issue, it should have been "tests". The rest of the changes are around making the types fail compilation for this mistake.

rwjblue:
I'd like to get more tests landed, but unsure exactly how best to go about it.

I could see a couple different ways to do it:

* add some unit tests for `splitter`
* add a new package that is a "test only" package

rwjblue:
Currently testing in `@ember/test-helpers` with https://github.com/emberjs/ember-test-helpers/pull/963.

ef4:
I happen to be writing unit tests for splitter right this second (for unrelated reasons). 

But adding a new test package and doing an end-to-end build and inspecting the build output to ensure that test code is only in the tests bundle would be good too.

PR #334, opened 2020-12-17
feat: resolve alias as prefix

buschtoens:
Webpack's [`resolve.alias`](https://webpack.js.org/configuration/resolve/#resolvealias) supports resolving prefix aliases.

```js
{
  resolve: {
    alias: {
      'react-query': 'react-query/es'
    }
  }
}
```

For instance the above config translates all paths starting with `react-query`:

- `react-query` ➡️ `react-query/es` (exact match)
- `react-query/foo` ➡️ `react-query/es/foo` (prefix match)
- `react-query-devtools` ➡️ `react-query-devtools` (no match, unchanged)

`ember-auto-import`'s `alias` option only supports exact matches so far.

This PR tweaks the `aliasFor(name)` method to also perform prefix matches in addition to exact matches.

### DIFF
--- a/packages/ember-auto-import/ts/package.ts
+++ b/packages/ember-auto-import/ts/package.ts
@@ -273,7 +273,14 @@ export default class Package {
   }
 
   private aliasFor(name: string): string {
-    return (this.autoImportOptions && this.autoImportOptions.alias && this.autoImportOptions.alias[name]) || name;
+    let alias = this.autoImportOptions?.alias;
+    if (!alias) return name;
+    if (alias[name]) return alias[name];
+
+    let prefix = Object.keys(alias).find(p => name.startsWith(`${p}/`));
+    if (prefix) return alias[prefix] + name.slice(prefix.length);
+
+    return name;
   }
 
   get fileExtensions(): string[] {
--- a/packages/sample-direct/app/templates/components/hello-world.hbs
+++ b/packages/sample-direct/app/templates/components/hello-world.hbs
@@ -1,5 +1,6 @@
 <div class="hello-world">{{formattedDate}}</div>
 <div class="lodash">{{#if lodashPresent}}yes{{else}}no{{/if}}</div>
 <div class="aliased">{{aliasedResult}}</div>
+<div class="prefix-aliased">{{prefixAliasedResult}}</div>
 <div class="scoped">{{fromScoped}}</div>
 <div class="module-dependency">{{moduleDependency}}</div>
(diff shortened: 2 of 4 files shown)


packages/sample-direct/app/templates/components/hello-world.hbs
```diff
@@ -1,5 +1,7 @@
 <div class="hello-world">{{formattedDate}}</div>
 <div class="lodash">{{#if lodashPresent}}yes{{else}}no{{/if}}</div>
 <div class="aliased">{{aliasedResult}}</div>
+<div class="aliased">{{aliasedResult}}</div>
```
buschtoens, line 4:
```suggestion
```

packages/ember-auto-import/ts/package.ts
```diff
@@ -273,7 +273,15 @@ export default class Package {
...
 
   private aliasFor(name: string): string {
-    return (this.autoImportOptions && this.autoImportOptions.alias && this.autoImportOptions.alias[name]) || name;
+    let alias = this.autoImportOptions?.alias;
+    if (!alias) return name;
+
+    for (const [prefix, replacement] of Object.entries(alias)) {
+      if (prefix === name) return replacement;
+      if (name.startsWith(`${prefix}/`)) return replacement + name.slice(prefix.length);
+    }
```
buschtoens, line 282:
```suggestion
    if (!alias) return name;
    if (alias[name]) return alias[name];

    let prefix = Object.key(alias).find(p => name.startsWith(`${p}/`));
    if (prefix) return alias[prefix] + name.slice(prefix.length);
```

packages/ember-auto-import/ts/package.ts
```diff
@@ -273,7 +273,14 @@ export default class Package {
   }
 
   private aliasFor(name: string): string {
-    return (this.autoImportOptions && this.autoImportOptions.alias && this.autoImportOptions.alias[name]) || name;
+    let alias = this.autoImportOptions?.alias;
+    if (!alias) return name;
+    if (alias[name]) return alias[name];
+
+    let prefix = Object.key(alias).find(p => name.startsWith(`${p}/`));
```
buschtoens, line 280:
```suggestion
    let prefix = Object.keys(alias).find(p => name.startsWith(`${p}/`));
```

ef4:
Thanks, this looks good.

ef4:
This [got reported as a breaking change](https://github.com/ef4/ember-auto-import/issues/368) in 1.11, I'm thinking we should add a flag to opt in to this behavior, and then make the flag default in 2.0.

PR #345, opened 2021-02-06
ease usage with ember-cli-content-security-policy

jelhan:
Currently all users of [Ember CLI Content Security Policy](https://github.com/rwjblue/ember-cli-content-security-policy) must explicitly set `forbidEval` to `true` in Ember Auto Import's configuration. Unless the very rare case that they allow `"unsafe-eval"` in `style-src` directive.

This is disruptive for users of both sides. Since Ember CLI blueprints for new application includes Ember Auto Import and many addons depend on it as well nearly all user of Ember CLI Content Security Policy must set that configuration. It does not help with adoption of CSP or testing for CSP compliance in Ember ecosystem.

This pull request propose to change the default value to depend on the presence of ember-cli-content-security-policy depenendency. If the application depends on ember-cli-content-security-policy `forbidEval` defaults to `true`. Otherwise it is `false`.

I would not consider this as a breaking change. At least I'm not aware of a meaningful use case to not having `forbidEval` set to `true` if CSP is used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ Supported Options
 
  - `alias`: _object_, Map from package names to substitute packages that will be used instead.
  - `exclude`: _list of strings, defaults to []_. Packages in this list will be ignored by ember-auto-import. Can be helpful if the package is already included another way (like a shim from some other Ember addon).
- - `forbidEval`: _boolean_, defaults to false. We use `eval` in development by default (because that is the fastest way to provide sourcemaps). If you need to comply with a strict Content Security Policy (CSP), you can set `forbidEval: true`. You will still get sourcemaps, they will just use a slower implementation.
+ - `forbidEval`: _boolean_, controls if `eval` is used in development. It is used because that is the fastest way to provide sourcemaps. Defaults to `false` unless [ember-cli-content-security-policy](https://github.com/rwjblue/ember-cli-content-security-policy) is used.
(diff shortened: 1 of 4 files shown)


jelhan:
Not fully sure why `dependencySatisfies` macro from `@embroider/macros` fails. But before digging deeper into it I would appreciate some feedback if you are open to accept such a change.

ef4:
Thanks, I like this.

`dependencySatisfies` only asks about your own dependencies, because you can only reliably access your own dependencies (without depending on accidental implementation details that we don't want to lock in forever). 

In a case like this one, `ember-auto-import` can declare an [optional peer dependency](https://github.com/emberjs/rfcs/blob/4be2d5b4a34b51e31eef6c6132dda0ef42495354/text/0507-embroider-v2-package-format.md#optional-peer-dependencies) on `ember-cli-content-security-policy`. The benefit of requiring this is that it makes it clear to package managers that you want access to the same copy of that library that is present in your parent package, if it happens to be there.

jelhan:
I added `ember-cli-content-security-policy` as an optional peer dependency in 29d4183f715c8ce2ec897d7c7e5c54c01a7bd621. But tests are still failing. Not sure if I done something wrong. But I'm wondering if `@embroider/macros` babel plugin is running at all. I would have expected it to fail when running the babel plugin. And not silently skipping the code and than throwing at execution time.

I can refactor to classical methods to determine if the consuming application has the dependency. I was mostly using `@embroider/macros` to demonstrate the concept. Haven't checked if this specific usage is supported at all, if targeted Ember versions are matching, etc.

jelhan:
@ef4 Would it be an acceptable path forward for you to _not_ use `@embroider/macros` but a classic strategy like [ember-cli-version-checker](https://github.com/ember-cli/ember-cli-version-checker)?

PR #347, opened 2021-02-08
WIP: Support importSync from @embroider/macros

simonihmig:
Fixes https://github.com/embroider-build/embroider/issues/684

@ef4 open questions:
* I am assuming that `importSync()` is expected to be used only with a string literal (or static template literal), is that right? I.e. I don't need to care about expressions in a template literal, as is the case for `import()`, do I?
* Besides the unit test, I tested this successfully locally in a real addon. And so yes, the analyzer runs before the macros transform, which is good. However we should add some integration-like test I think, just unsure how you would want to set this up in this project. Should I add a new `sample-embroider-macros` package? Or would our test case fit into any of the existing ones?


ef4:
Thanks.

This should accept exactly the same things `import()` does, I think it can share most of the implementation. 

As for tests, I think it’s fine to add an end-to-end test of this feature to sample-direct.

simonihmig:
> This should accept exactly the same things import() does, I think it can share most of the implementation.

Oh, really? Does Embroider support template strings with dynamic expressions for this?

I was looking at the [importSync docs](https://github.com/embroider-build/embroider/tree/master/packages/macros#importsync) saying this...

> importSync takes a single static string as its only required argument.

If that is the case indeed, it *seems* to me the changes required are somewhat more involving. Looking at the `TemplateImport` interface and glimpsing over `splitter.ts` (especially [`BundleDependencies`](https://github.com/ef4/ember-auto-import/blob/master/packages/ember-auto-import/ts/splitter.ts#L42-L46)), it seems to imply a dynamic import, doesn't it? (it has no `isDynamic: boolean` prop as `LiteralImport`). But that means it will get it's own code-splitted chunck, which is not what `importSync()` should do, right?

ef4:
>  Does Embroider support template strings with dynamic expressions for this?

Yes, it does. 

That doesn't imply the code will go into a new chunk though. Since it's synchronous, all matching modules will get built into the current bundle, but will only execute when asked for.

But you don't need to implement all that yourself. If you copy the existing template for dynamic template imports:

```
{{#each dynamicTemplateImports as |module|}}
    d('_eai_dynt_{{js-string-escape module.key}}', [], function() {
      return function({{module.args}}) {
        return import({{{module.template}}});
      }
    });
{{/each}}
```

And replace `import()` with `require()`, webpack will do the right thing.

simonihmig:
I pushed some changes, including some test cases in `sample-direct`.

The string literal version still works, but the template does not. And I am not quite getting how this should be made to work...

[This](https://github.com/ef4/ember-auto-import/pull/347/files#diff-b41cd505c62380edeffd814f9c2ff2382c66758aef8b8dbc6948c4d5c806a953R35) case of a template string import gets compiled to:

```js
    templateDependency: Ember.computed('which', function () {
      var _importSync6 = require('another-dependency/flavors/' + (this.get('which') || 'vanilla')),
          name = _importSync6.name;

      return name;
    })
```

But the `require()` call won't find that module, as it never has been defined. As you suggested (hope I got you right), the template-based imports get defined in the entrypoint like [this](https://github.com/ef4/ember-auto-import/pull/347/files#diff-8b109a24d021f0efea4c42a6a90ea5afa9f93ff20f7d6528cea5d7f4ddd35a60R54-R60), similar to the template-based dynamic imports. But this doesn't work here, as unlike `import()` the `importSync()` call does not get transpiled by e-a-i's own babel plugin into those `emberAutoImportDynamic()` calls. 

So how could that work here? Do we have to extend the babel plugin to also transpile those? But that means we are transpiling every `require()`, which we don't want, right? Or should we do our own `importSync()` handling, by transpiling it to `emberAutoImportDynamic` calls *before* `@embroider/macros` does that?

And there is no way to make all these possible modules given by a dynamic template string into static `d()` calls like a static import, at least without duplicating what webpack does there, right?

Again, all of that is no problem if we restrict `importSync()` to string literals. Given that, by coincidence I stumbled upon [this line](https://github.com/embroider-build/embroider/blob/master/packages/core/src/babel-plugin-adjust-imports.ts#L78) in `embroider/core`, which does suggest that only string literals are supported, not?

Hope you can clear up my confusion! :wink:

ef4:
> Given that, by coincidence I stumbled upon this line in embroider/core, which does suggest that only string literals are supported, not?

No, that is only restricting which cases the adjust-imports-plugin handles. It means that certain compatibility features -- like the ability for an addon to expose modules under a different package name -- don't work with `importSync`.

>  Or should we do our own importSync() handling, by transpiling it to emberAutoImportDynamic calls before @embroider/macros does that?

Yes, I think this is the way to go.

simonihmig:
> Yes, I think this is the way to go.

So my current WIP babel plugin extension does this now quite well... 

*However*:

1. the user has to opt in currently into eai's custom import babel plugin. This then would also apply to being able to use `importSync()` (at least with a template string). I am fine with that, just want to double check!

2. Unfortunately now, eai's babel plugin seems to always run *after* that of `@embroider/macros`, so only sees that generic `require()` which we cannot use as the basis for own transformations. I am not aware of any way to enforce a specific order with `ember-cli-babel`, in the sense of running plugin A explicitly before plugin B. Are you? /cc @rwjblue 

The only other somewhat hacky solution that came to my mind was to allow some kind of orchestration between eai and macros - which, given that they are aware of each other and eai will by this PR get even more coupled to macros, might be ok!? 

Something like letting macros add a hint by transforming `importSync(`module/${what}`)` to `require(/* importSync */ `module/${what}`)`, which eai's babel plugin then can distinguish from ordinary `require()` calls and apply its transformation to `emberAutoImportDynamic()`. 🤔

ef4:
Babel plugin ordering is a little complicated to reason about.

Plugins run in the order they're configured, but *per AST node*. So a later-configured plugin that manipulates higher AST nodes will still do its thing before an earlier-configured plugin that manipulates deeper AST nodes.

simonihmig:
Yeah, I read about it, and got it. But in this case two plugins are competing for the *same* AST node. 

So we either have to enforce a certain order (and idk how?), or we must make it work in both possible orders (which is basically what I suggested with that comment hint trick).

ef4:
Here's an idea you can try, I *think* it will work: assuming both our plugins want to manipulate `Import` nodes, we can make them collaborate by making one work on `enter` and one work on `exit`.

This:

```js
visitor: {
      Import(path: NodePath<Import>) {
      }
}
```

is shorthand for:

```js
visitor: {
      Import: {
        enter(path: NodePath<Import>) {
        }
      }
}
```

And you can also add an exit:

```js
visitor: {
      Import: {
        exit(path: NodePath<Import>) {
        }
      }
}
```

I guess we want auto-import to go first, so it would continue working on `enter`, and we would change macros to work on `exit`.

simonihmig:
Ok, cool, i'll try that!

simonihmig:
@ef4 this works! As soon as https://github.com/embroider-build/embroider/pull/694 is merged and released, and we update the dependency here, tests *should* turn green here. Were passing locally with `yarn link`.

Another thing though: I added a new test-package `another-dependency` here (which does the same as `a-dependency`), for the reason that `sample-direct` was doing `import('a-dependency')`, so I was unsure if it's a reasonable thing to *also* synchronously import the same package. 

And indeed this causes problems, when I removed `another-dependency` and tried to both ``import(`a-dependency/flavor/${...}`)`` and ``importSync(`a-dependency/flavor/${...}`)``. This is how the app.js looks like when compiled:

image

The first one comes from `importSync`, and webpack turns that correctly into a *sync* dependency, while the second is from dynamic import, and so a lazy/async dependency.

But the `emberAutoImportDynamic` that `importSync()` compiles to will use the later, unexpectedly returning a Promise, which makes things fail. 

So, what to do? It wouldn't be hard to properly separate the two, like having a `_eai_sync_a-dependency/flavors/${e}'` module and a separate `window.emberAutoImportSync()` to consume that, so both ways to refer to that (*same*) module don't collide.

However that means you basically get the same module once in you main bundle, and once in a splitted chuck, which doesn't make too much sense. Certainly we can (and maybe should?) leave that to the user to optimize/get right, but I wondered if eventually the analyzer should detect that and handle somehow? (throw?) 

The same basically applies when you do `import a from 'a-dependency'` *and* `await import('a-dependency')`, and AFAICT that case is also not handled in any special way. So I guess that's not what we want to do, but I just wanted to double-check with you first!

simonihmig:
@ef4 I updated this to the recent new release of macros, so CI is happy. 

However the issue in the previous comment still exists, could you provide a direction here?
* either we make synchronous (`importSync()`) and asynchronous (`import()`) imports coexist peacefully, with the caveat that the user will get the same module twice
* or we try to detect that before and throw an error

ef4:
Yes, you're right and we should register the sync and async versions under different names (`_eai_sync_a-dependency`).

I don't actually think webpack will duplicate the underlying library in that case. It's OK if we import it twice, in two different styles. That's a thing webpack already needs to get right.

simonihmig:
>  we should register the sync and async versions under different names (_eai_sync_a-dependency).

Did that, and updated the tests for the latest test infrastructure changes on master. @ef4 ready for final review!

ef4:
Thanks @simonihmig, great work here.

PR #350, opened 2021-02-16
introduce watchDependencies option

ef4:
This lets you rebuild in development when editing an auto-imported package.

### DIFF
--- a/packages/ember-auto-import/ts/webpack.ts
+++ b/packages/ember-auto-import/ts/webpack.ts
@@ -114,7 +114,9 @@ export default class WebpackBundler implements BundlerHook {
       },
       output: {
         path: this.outputDir,
-        filename: `chunk.[id].[chunkhash].js`,
+        // entry chunks need to have stable names, so we can more easily gather
+        // them all up to append to Ember's vendor.js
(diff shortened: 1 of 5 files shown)


packages/ember-auto-import/ts/webpack.ts
```diff
@@ -163,10 +171,6 @@ export default class WebpackBundler implements BundlerHook {
       use: {
         loader: 'babel-loader-8',
         options: {
-          // do not use the host project's own `babel.config.js` file
-          configFile: false,
-          babelrc: false,
```
rwjblue, line 168: Why are these removed? As far as I can tell, we need them still (or we run much more than `@babel/preset-env` on deps that are imported).

rwjblue: This directly reverts the fixes landed in https://github.com/ef4/ember-auto-import/pull/342... 🤔

rwjblue: Submitted https://github.com/ef4/ember-auto-import/pull/360 to bring back the changes from #342. Happy to discuss over there...

Issue #354, opened 2021-02-24
Somehow detect app namespace import


lifeart:
given:

yarn powered ember project, named `sed`.

in project, widely used project namespace imports like

```js
import model from 'sed/models/model';
```

project has:

package.json
```json
{
   "name": "sed",
   "dependencies": {
     "sed": "file:app"
   }
}
```

if we add `ember-auto-import` into project, it's going crazy, because it's trying to resolve project, using webpack, because this checks it true:

`dependency in package.json`, it does not have `ember-addon` keyword in package.json (`!isEmberAddonDependency`).

what to do?

check root.package.json[name], and if name itself in deps, ignore it or wrarn.

ef4:
I think it's pretty weird for a project to list itself in `dependencies`.

Maybe you can get the same effect using the newer [Package Entry Points](https://nodejs.org/dist/latest-v14.x/docs/api/all.html#packages_package_entry_points) options? I think it would let you say:

```js
{
  "name": "sed",
  "exports": {
    "./*": "./app/*.js",
  }
}
```

lifeart:
I agree on weird case, I suppose it was created to get native imports autocomplete working, and without auto-import it working just fine (in my understanding this hack boy used in build process). But, I found debug for this case quite tricky, and wondering, should we prevent leg-shot on auto-import level (and may be embroider)

ef4:
Yes, it’s probably a good idea for auto-import to always ignore the app’s
own package name.

(Making subdirectories of the app auto-importable is actually a pretty nice
feature I’m open to adding, but it would be explicitly opt in.)

ef4:
I don't think we should support apps that list themselves in dependencies.

PR #355, opened 2021-02-25
handle symlinked app case

lifeart:
resolves https://github.com/ef4/ember-auto-import/issues/354


ef4:
I don't think we should support apps that list themselves in dependencies.

PR #362, opened 2021-03-18
Fix compatibility with ember-cli-babel@7.26+

### LINKED ISSUE #359: Dependencies in `.ts` files not included with latest `ember-cli-babel`
It looks like `ember-cli-babel@7.24.0` [dropped several items it previously returned](https://github.com/babel/ember-cli-babel/commit/5eafd6baad763786246e06f862704a770be6ffc9#diff-4d3a8edb07aff5a7c43d8692de88a4eb83b327e2a352633699ba4afb9b753b6aL39-L51) from `getBabelOptions()`, including the `filterExtensions` array that `auto-import` was relying on to determine whether to analyze `.ts` files.

It seems like we either need to duplicate `ember-cli-babel`'s "should we process TypeScript" logic here, or (I think preferably) `ember-cli-babel` should expose an explicit way to retrieve its configured file extensions.

I think this is likely the cause of the recently added comments on https://github.com/ef4/ember-auto-import/issues/267

rwjblue:
ember-cli-babel@7.24 fixed a bug where the `buildBabelOptions` result included a number of non-babel specific properties (reported in https://github.com/babel/ember-cli-babel/issues/227). Unfortunately, while we were explicitly working around that bug (the `delete`'s in the snippet here) but we **also** ended up relying on it with the usage of `filterExtensions` property.

In order to make it possible for ember-auto-import to have a path forward, ember-cli-babel@7.26.0 introduces a new public API to get a list of the supported extensions (`babelAddon.getSupportedExtensions`).

This PR uses the presence of the `getSupportedExtensions` function to decide if it needs to clear the abnormal properties or not, and fixes usage of ember-auto-import along with ember-cli-babel@7.26+.  

Fixes #359
Fixes https://github.com/babel/ember-cli-babel/issues/389


packages/ember-auto-import/ts/package.ts
```diff
@@ -127,17 +127,26 @@ export default class Package {
     // is using. We will use these so we can configure our parser to
     // match.
     let babelAddon = instance.addons.find(addon => addon.name === 'ember-cli-babel') as any;
-    let babelOptions = babelAddon.buildBabelOptions(options);
-    let extensions = babelOptions.filterExtensions || ['js'];
+    let version = parseInt(babelAddon.pkg.version.split('.')[0], 10);
+    let babelOptions, extensions;
+
+    if (typeof babelAddon.getSupportedExtensions === 'function') {
+      babelOptions = babelAddon.buildBabelOptions('babel', options);
```
rwjblue, line 134: I added this new signature so that I could make the default invocation without arguments (or options) passed return the broccoli-babel-transpiler style options.

ef4:
Now that the test updates have landed, we can probably just update the ember-cli-babel version in `app-template` to the new one, and add a new test scenario that explicitly covers the old one. The old one can become an aliased dependency in the `test-scenarios` package.

ef4:
This is breaking ember-auto-import 2.0.0-beta.1 due to https://github.com/babel/ember-cli-babel/issues/401.

I'm going to revert it and rely on the backward compatible call signature instead.

Issue #365, opened 2021-03-24
v1.11.0 causes webpack SplitChunksPlugin error


timiyay:
Dependabot opened an update PR against several of our apps for `ember-auto-import@v1.11.0`, and our CI is failing with the following error:

```
/app/node_modules/webpack/lib/optimize/SplitChunksPlugin.js:803
								throw new Error(
								^

Error: SplitChunksPlugin: You are trying to set a filename for a chunk which is (also) loaded on demand. The runtime can only handle loading of chunks which match the chunkFilename schema. Using a custom filename would fail at runtime. (cache group: vendors)
    at /app/node_modules/webpack/lib/optimize/SplitChunksPlugin.js:803:15
    at SyncBailHook.eval (<anonymous>:7:16)
    at SyncBailHook.lazyCompileHook (/app/node_modules/tapable/lib/Hook.js:154:20)
    at Compilation.seal (/app/node_modules/webpack/lib/Compilation.js:1336:38)
    at /app/node_modules/webpack/lib/Compiler.js:675:18
    at /app/node_modules/webpack/lib/Compilation.js:1261:4
    at AsyncSeriesHook.eval [as callAsync] (<anonymous>:15:1)
    at AsyncSeriesHook.lazyCompileHook (/app/node_modules/tapable/lib/Hook.js:154:20)
    at Compilation.finish (/app/node_modules/webpack/lib/Compilation.js:1253:28)
    at /app/node_modules/webpack/lib/Compiler.js:672:17
    at _done (<anonymous>:9:1)
    at eval (<anonymous>:20:22)
    at /app/node_modules/webpack/lib/Compilation.js:1185:12
    at /app/node_modules/webpack/lib/Compilation.js:1097:9
    at processTicksAndRejections (internal/process/task_queues.js:75:11)
```

This error is occurring when one of our addons uses dynamic imports load `Intl` polyfills for `ember-intl`, like so:

```js
await import('@formatjs/intl-getcanonicallocales/polyfill')
await import('@formatjs/intl-locale/polyfill')
await import('@formatjs/intl-pluralrules/polyfill')
await import('@formatjs/intl-numberformat/polyfill')
await import('@formatjs/intl-relativetimeformat/polyfill')
await import('@formatjs/intl-datetimeformat/polyfill')
```

Turbo87:
same issue for us:

```
[...]

PR #367, opened 2021-03-24
Fixing common chunk bug

ef4:
This reproduces issue #365.


ef4:
We were trying to configure webpack to always have stable names for entry chunks, so that we wouldn't need to do any cleanup of stale chunks in the step where we're figuring out what to append into ember's vendor.js. But the configure we tried makes SplitChunksPlugin unhappy when there are lazy common chunks. So I'm dropping that config and making our own build do cleanup of stale chunks.

PR #370, opened 2021-03-24
Reimplement prefix aliasing

ef4:
@buschtoens here is an open PR for the prefix matching feature that I had to revert. I left a failing test for the remaining bit.


ef4:
Closing in favor of the work in 186f34c4204896dc93948037fb6b26fd31ae2564

PR #379, opened 2021-04-06
2.0 upgrade guide

ef4:
This is the plan for how I want to roll out 2.0. Not everything described in here is implemented at the moment, but it's close.

 - [x] [webpack 5](https://github.com/ef4/ember-auto-import/pull/327)
 - [x] [embroider v2 addon format](https://github.com/ef4/ember-auto-import/pull/376)
 - [x] [drop babel 6](https://github.com/ef4/ember-auto-import/pull/382)
 - [x] [inject entry chunks directly in html](https://github.com/ef4/ember-auto-import/pull/385)
 - [x] [assert mandatory top-level import, with semver-matching leader election](https://github.com/ef4/ember-auto-import/pull/383)
 - [x] [app responsible for installing webpack](https://github.com/ef4/ember-auto-import/pull/389)
 - [x] change alias option to match webpack's prefixing rules


docs/upgrade-guide-2.0.md
```diff
@@ -0,0 +1,54 @@
...
+ember-auto-import 2.0 upgrades from webpack 4 to 5. This is a potentially breaking change for any package that directly adds to webpack config. Most apps will not experience any breakage, because most common webpack 4 configs still work in webpack 5.
+
+### Embroider v2 Addon Format support
+
+ember-auto-import 2.0 handles Embroider v2-formatted addons for apps that aren't yet using Embroider.
+
+This is potentially breaking because it adds more stuff to the webpack config that people could have been manually controlling before. Specifically:
+
+- v2 addons are allowed to import css, so we add the appropriate webpack loaders
+- v2 addons are allowed to contain inline templates, decorators, and embroider macros, so all of these are now supported by ember-auto-import
```
buschtoens, line 23: This is _huuuuge!_ 😍

If I'm understanding this right, it means that we can iteratively make our individual in-house addons / engines Embroider-v2-formatted-compatible, without having to also fully migrate the host app in the same swoop, but still reap the build performance improvements by pre-building & publishing the addons / engines to a registry.

Is that correct?

And if so, are there recommended workflows for monorepo setups already? I would like to use the pre-built packages, if there are no local changes, and only live-build them, if changes were made.

ef4: Yes, that's correct. Addons will be able to port to v2. The only requirement that will impose on apps will be that they take ember-auto-import 2.0.

> And if so, are there recommended workflows for monorepo setups already? I would like to use the pre-built packages, if there are no local changes, and only live-build them, if changes were made.

I don't have any examples yet of that kind of workflow, but I would expect that people wouldn't want to commit build artifacts to their repo, so you'd have a dev process analogous to running `tsc --watch` that could cover the whole monorepo. It's not going to be expensive, because the kind of transformations you'd do (stripping typescript types, inlining templates, generating `setComponentTemplate`) are quite light.

Issue #169, opened 2018-12-11
Error importing type definitions?


kernel-io:
If you find you are getting an error while trying to import type definitions, make sure to add the name of the package with type definitions to the excludes array in your autoImport configuration in ```ember-cli-build.js``` :+1:

kernel-io:
I'm not educated regarding the benefits of using the babel plugin for analysis, however if using it brings more benefits / opens up more avenues than option 1, wouldn't it be better to follow that path? Option 1 seems very very specific to this issue only?

tschoartschi:
We actually hit the same problem... But we can not work around this issue because the package has typings and source in the same package. What we would need is:

```
autoImport: {
    exclude: ['web-sdk/lib/definitions']
}
```

```typescript
import WebSdk from 'web-sdk'; // behind this real JS code sits
import {GlbViewer} from 'web-sdk/lib/definitions/glb-viewer-core/src/glb-viewer'; // this is only type information

export const GLB_CANVAS_ID = 'glb-canvas-container';

export default class GLBViewer extends Component {
    private _glbViewer: GlbViewer;

    constructor() {
    	this._glbViewer = await WebSdk.getGlbViewer();
    }
}    
```

Since the package `web-sdk`  is a package we control we could adjust it, if it makes sense but many customer use it already so we need a good reason to change it. There where no problems with any of the other major frameworks (Vue, React, Angular, custom Rollup build).

Would be great if we could provide a workaround for Ember users

tschoartschi:
@dfreeman showed us a workaround in discord:

```
export default class GLBViewer extends Component {
    private _glbViewer: import('web-sdk/lib/definitions/glb-viewer-core/src/glb-viewer').GlbViewer;
}
```
But that's not very nice it looks awkward, furthermore the IDE always suggests the former import style. So we always have to manually adjust

ef4:
There is another case that I think is worth including in this issue. When importing something _only_ for use as a type, you really don't want that to create a real runtime dependency on the module. That interferes with your ability to lazily load the module.

Moving our analysis into a babel plugin isn't even a slam dunk, because there would still be finicky timing concerns there. Changes to the implementations of other babel plugins could break the analysis.

So I'm starting to think we need to introduce our own detection of type-only imports.

boris-petrov:
Is this bug going to be circumvented when using the soon-to-be-available [import type](https://github.com/babel/babel/issues/10981) functionality of TypeScript?

[...]

PR #380, opened 2021-04-08
fix(analyzer): ignore type-only imports & exports

buschtoens:
Prevents [type-only imports / exports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export) from ending up in the final bundle.

```ts
import type { compute } from 'heavy-library';

export async function performHeavyComputation(
  ...args: Parameters<typeof compute>
): Promise<ReturnType<typeof compute>> {
  const { compute } = await import('heavy-library');
  return compute(...args);
}
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -828,10 +828,12 @@
       }
     },
     "node_modules/@babel/helper-plugin-utils": {
-      "version": "7.13.0",
-      "resolved": "https://registry.npmjs.org/@babel/helper-plugin-utils/-/helper-plugin-utils-7.13.0.tgz",
-      "integrity": "sha512-ZPafIPSwzUlAoWT8DKs1W2VyF2gOWthGd5NGFMsBcMMol+ZhK+EQY/e6V96poa6PA/Bh+C9plWN0hXO1uB8AfQ==",
-      "license": "MIT"
+      "version": "7.14.5",
+      "resolved": "https://registry.npmjs.org/@babel/helper-plugin-utils/-/helper-plugin-utils-7.14.5.tgz",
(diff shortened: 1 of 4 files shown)


ef4:
Thanks for working on this, this will be nice.

boris-petrov:
Does this fix [this issue](https://github.com/ef4/ember-auto-import/issues/169#issuecomment-603088345) (i.e. this comment in that issue)?

ef4:
Yes, I think this would fix that.

jamescdavis:
Bump. 🙂

buschtoens:
Alright, sorry that I kept this idling for so long, but I had no clue and time to dig into, why the 4 `*-own` scenarios were failing, while all other were running successfully. I finally got around to it now!

The issue was that _all_ scenarios, except for `*-own`, seem to directly or indirectly enable the TypeScript syntax in the babel options. Without the TS syntax enabled, TS constructs like `import type ...` cannot be parsed, which is obvious in hindsight, but was quite annoying to debug, as there was no apparent syntax error getting thrown. 😅

I've added `@babel/plugin-syntax-typescript` to the `babelOptions` in the `beforeEach` hook via 144005204bef33779fc7720dc2fe63b6e2971fa3, which fixes the issue. I don't really know, whether this solution is acceptable / ideal though, or whether I should opt-in to TS somewhere else.
Any pointers would be greatly appreciated!

I've also rebased the PR and it's all working fine now. ✅

Please let me know, whether I should add more tests, or whether I should change the TS syntax opt-in.

ef4:
Looks great, thanks.

Issue #261, opened 2019-12-16
Dynamically import CSS


boris-petrov:
I'm trying to achieve the same functionality as dynamically-importing JavaScript (in order to have a smaller bundle size). That is, I have a CSS file that I want to load on-demand. What I do right now is:

ember-cli-build.js:
```javascript
new Funnel('node_modules/c3', { destDir: '/assets', files: ['c3.css'] }),
```

app code:
```typescript
function importCss(url: string): void {
  const link = document.createElement('link');
  link.rel = 'stylesheet';
  link.type = 'text/css';
  link.href = url;
  document.getElementsByTagName('head')[0].appendChild(link);
}

importCss('/assets/c3.css')
```

Not the best code as you see. I mostly hate the part in `ember-cli-build`. I guess `ember-auto-import` can help mostly with it. I read [this](https://github.com/ef4/ember-auto-import/issues/66#issuecomment-405328064) answer - point 2 seems like something similar to what I need but it doesn't work - I think the browser tries to parse the result file as JavaScript. [This](https://github.com/ef4/ember-auto-import/pull/205) PR also doesn't do what I want.

Any suggestions as to what I could do or what should be developed in `ember-auto-import` in order to support that?

ef4:
I think #205 supports your use case, but you need to bring your own webpack config for it.

```js
let app = new EmberApp(defaults, {
  autoImport: {
    webpack: {
      // In here you need to configure something like webpack's style-loader
    }
  }
});
```

boris-petrov:
That PR suggests that the included CSS files will end up in `vendor.js` which I *don’t* want. I want them to be separate files, same as the JS chunk files created. And then, perhaps, have an easy way of importing them from the frontend (that’s not mandatory - the current way is almost OK).

ef4:
Whether the code ends up in vendor vs another bundle is determined by whether anybody is dynamically importing it. Try adding style-loader and do `await import('./your/style.css')`. I think it may just work.

boris-petrov:
@ef4 - thanks for the support. So here is what I did:

```javascript
...
module: {
  rules: [
    {
      test: /\.css$/i,
      use: [
        { loader: 'style-loader', options: { injectType: 'linkTag' } },
        'css-loader',
      ],
    },
  ],
},
...
```

This in the `webpack` part of `autoImport` in `ember-cli-build`. Then I have `await import('c3/c3.css');`.

[...]

PR #408, opened 2021-05-26
Add ability to specify options to `style-loader` and `css-loader`

boris-petrov:
This PR concerns [that](https://github.com/ef4/ember-auto-import/issues/261#issuecomment-847613363) comment.

I'm not sure I've done it right, nor am I sure whether this is the correct way to approach things. @ef4 - please tell me if you think a completely different thing should be done.


ef4:
Thanks.

PR #414, opened 2021-06-15
[BUGFIX] webpack detection should also consider supported versions

stefanpenner:


packages/ember-auto-import/ts/auto-import.ts
```diff
@@ -98,16 +100,14 @@ export default class AutoImport implements AutoImportSharedAPI {
...
-    } catch (err) {
-      if (err.code !== 'MODULE_NOT_FOUND') {
-        throw err;
-      }
-      throw new Error(`this version of ember-auto-import requires the app to have a dependency on webpack 5`);
+    } else {
+      throw new Error(`[ember-auto-import] this version of ember-auto-import requires the app to have a dependency on webpack 5`);
     }
-
+
```
thoov, line 110: Extra whitespace

thoov (approved):

stefanpenner:
one of the runs failed due to what appears to be an npm issue (unrelated to thsi PR)

stefanpenner:
the new failures also look spurious. One looks suspicious but doesn't appear to be reproducible locally https://github.com/ef4/ember-auto-import/pull/414/checks?check_run_id=2833013788

Should I just re-roll the CI dice?

ef4:
I think those fastboot failures are real and probably caused by upstream changes in a fastboot-related package. You may not see them locally if you're installing via lockfile.

stefanpenner:
> I think those fastboot failures are real and probably caused by upstream changes in a fastboot-related package. You may not see them locally if you're installing via lockfile.

I'll try that

stefanpenner:
@ef4 yup, lockfile prevented local reproduction. Thanks for pointing in that direction.

```
not ok 2 fastboot3-dev > it is using schema version 5
  ---
  message: "failed"
  severity: failed
  actual: 3
  expected: 5
  stack:     at Assert.equal (/Users/stefanpenner/src/ef4/ember-auto-import/node_modules/qunit/qunit/qunit.js:3072:14)
  ```

File: test-scenarios/fastboot3-test.ts [hunk 2]
```diff
@@ -22,7 +22,7 @@ function upgradeFastbootFormat() {
   let upgrader = new Project('fastboot_upgrader', {
     files: {
       'index.js': `
-        const Funnel = require('broccoli-funnel');
+        const { Funnel } = require('broccoli-funnel');
```
stefanpenner, line 25: broccoli-funnel was not pinned, rather it was a random transitive dependency that was being pulled in. The latest version changed it's API, and decoupled its "functional" api from its "classical" api. (thanks @rwjblue)

To correct this, I explicitly added broccoli-funnel as a devDep + imported it correctly

stefanpenner:
bah, failure due to NPM install... rerunning

stefanpenner:
green! @ef4 I don't have release bit on NPM, so I'll leave that to you :)

PR #421, opened 2021-07-07
Fix publicAssetURL typo in v2 upgrade docs

kiwi-josh:
Could possibly be the cause of this issue (if people copy/pasta'd)?
https://github.com/getsentry/sentry-javascript/issues/3670


ef4:
Oof, thanks.

Issue #422, opened 2021-07-09
Outputs file when not used in app


mydea:
While trying to debug https://github.com/ef4/ember-auto-import/issues/420, I noticed that I always get a bundle (e.g. /assets/chunk.143.f3770ae5a87436ec20ec.js), even if auto-import is not used in my app.

Since ember-auto-import comes with the default blueprint and is needed for ember-qunit, it is not that uncommon to have it installed "as a side effect" and not because I actually use it in my app. In such cases, it seems unexpected to me to end up with an "unneeded" file:

```js
var __ember_auto_import__;
(() => {
    var e = {
            320: (e, r, t) => {
                var o "undefined" != typeof document && (t.p = (o = document.querySelectorAll("script"))[o.length - 1].src.replace(/\/[^/]*$/, "/")), e.exports = function() {
                    _eai_d
                    var e = _eai_r
                    window.emberAutoImportDynamic = function(r) {
                        return 1 === arguments.length ? e("_eai_dyn_" + r) : e("_eai_dynt_" + r)(Array.prototype.slice.call(arguments, 1))
                    }, window.emberAutoImportSync = function(r) {
                        return e("_eai_sync_" + r)(Array.prototype.slice.call(arguments, 1))
                    }
                }()
            },
            608: function(e, r) {
                window._eai_r = require, window._eai_d = define
            }
        },
        r = {}

    function t(o) {
        var n = r[o]
        if (void 0 !== n) return n.exports
        var _ = r[o] = {
            exports: {}
        }
        return e[o].call(_.exports, _, _.exports, t), _.exports
    }
    t.p = "", t(608)
    var o = t(320)
    __ember_auto_import__ = o
})()
```

Which adds a full HTTP request to my app.

I would expect to not have a file in production if I do not use ember-auto-import in my app code.

This can be seen on a "plain" app created with ember-cli 3.27, e.g. see: https://github.com/mydea/ember-issue-reproductions

Semi-related, I also noticed that it always outputs a chunk file for qunit, even when running `ember b --prod` or specifically setting `tests: false` in `ember-cli-build.js`. This has not necessarily an impact as the file is not referenced anywhere, but it is still deployed etc. along with the other files and also seems like it should not be generated when tests are not run.

Not sure if these issues are known/to be expected or not, but it did appear unexpected to me at least ;)

ef4:
I think this was addressed along with the changes in #424.

Issue #420, opened 2021-07-07
Error when auto import is not used


mydea:
I have an app that is not using ember-auto-import in its application code (as far as I can tell).

I noticed that I get a runtime error:

```
l.js:2 Uncaught ReferenceError: require is not defined
    at eval (l.js:2)
    at Object.../../../../../tmp/broccoli-20413LssqYisKTYOQ/cache-305-webpack_bundler_ember_auto_import_webpack/l.js (chunk.app.9a9a5f13351ef9d0df2b.js:29)
    at __webpack_require__ (chunk.app.9a9a5f13351ef9d0df2b.js:53)
```

Where `l.js` is:

```js
window._eai_r = require;
window._eai_d = define;
```

When I remove `ember-auto-import` from my devDependencies, the error goes away.

I would just remove it, but according to https://github.com/emberjs/ember-qunit/blob/master/docs/migration.md it is needed for ember-qunit for the tests (although the tests _do_ work without it, as far as I can tell :thinking: )

ef4:
What version of ember-auto-import?

What were you changing when this problem appeared?

mydea:
Not 100% sure, as the app did not fail (and tests passed), the error is only visible in the console (which I only just noticed). 

I would _guess_ it happend when upgrading from ember-auto-import 1.x to 2.x - I just tried downgrading it and the error went away!

I will try to see if I can get a reproduction somehow...!

mydea:
OK, after _a lot_ of trial and error, I found the culprit!

In this app, `defer` was added to the vendor JS file in index.html:

```html
<script src="{{rootURL}}assets/vendor.js" defer></script>
```

As it uses fastboot/prember and thus does not _need_ the JS to work. 

It works if:

* I remove `defer`
* _or_ I downgrade to ember-auto-import v1
* If I visit `/tests` it also does not trigger the error

Here is a reproduction repo:
https://github.com/mydea/ember-issue-reproductions/commit/4948624aa766b1d12edc067b545e0d096a630d13

ef4:
We need to find a clearer place to document this, but this is a known limitation, see https://github.com/ef4/ember-auto-import#customizing-build-behavior:

> `publicAssetURL`: the public URL to your /assets directory on the web. Many apps won't need to set this because we try to detect it automatically, but you will need to set this explicitly if you're deploying your assets to a different origin than your app (for example, on a CDN) or if you are using <script defer> (which causes scripts to be unable to guess what origin they loaded from).

[...]

Issue #409, opened 2021-05-26
Support changing the asset rootURL at runtime


ctcpip:
I can't compile after upgrading to v2.  Build is failing with:

```
Build Error (Inserter)

ember-auto-import could not find a place to insert app scripts in index.html.
```

ctcpip:
in my `index.html` I do not have hardcoded script/src tags.  Due to unique requirements of our application, the `src` path is modified dynamically at runtime.  So I think it is failing in that bit of code that is parsing the file and looking for a `<script>` with a `src` attribute.  Which it will never find.

ef4:
What do your `<script>` tags look like in index.html?

Can you share any customizations of `outputPaths` or `fingerprint` options in `ember-cli-build.js`?

Do you have a customized `rootURL` in `config/environment.js`?

Please run a build with the environment variable `DEBUG=ember-auto-import:*` and share the output.

ef4:
I replied above before seeing your second comment, yes, that must be the issue. Can you say more about why you need to do it that way? Maybe there's an easier alternative that would work with ember-auto-import, or maybe we need to make a more manual way for you to indicate where you want the auto-imported content to go.

ctcpip:
`"<script src=\"" + foo.bar + "{{rootURL}}assets/vendor.js\" charset=\"utf-8\"><\/script>"`

ctcpip:
without getting into too many gory details, the client needs to talk to the server to determine the root path due to requirements of an appliance in a datacenter with a reverse proxy...

ef4:
The challenge here is that we can produce some arbitrary number of bundles, depending on how many places your app and addons might say `await import()`. All of those bundles need this URL adjustment.

Some of those bundles will be entrypoint bundles and end up in script tags in index.html, and that is where this failure is coming from. We used to append the entrypoint bundles to vendor.js, but in 2.0 they became standalone files, which is better for caching and makes the build simpler.

And other bundles are lazy and will load only on demand, so you would also need to make sure webpack's runtime loader can find those. The way to do that is to [set the `__webpack_public_path__` global variable](https://webpack.js.org/guides/public-path/#on-the-fly). Even on ember-auto-import 1.x, you need to be careful because if anybody (including your addons) uses `await import()` you'll get a lazy bundle that won't load correctly in production if you don't set that.

[...]

PR #424, opened 2021-07-20
Support Customized HTML Insertion and Custom Entrypoints

### LINKED ISSUE #418: Support Custom Script Attributes when inserting
https://github.com/ef4/ember-auto-import/blob/1a581312739e1be8c0fc2426586df14a1a77d864/packages/ember-auto-import/ts/inserter.ts#L111

I have a use-case that needs to add a placeholder for SSR vs non-SSR `defer` attributes (replacement of the placeholder is handled by another library). 

Because our `defer` placeholder is being conditionally replaced in a build step _after_ ember-auto-import has run, the `publicAssetURL` [from the config](https://github.com/ef4/ember-auto-import#customizing-build-behavior) does not help us here.

I'm requesting that an official API be added for either customizing the attributes added to the script tags, or for replacing the default insertion wrapping string/function.

ef4:
Allow the customization of inserted HTML via custom elements. Among other things, planning to fix #418, #420, #422, and #409 here.

Adds two new options:

```diff
export interface Options {
  exclude?: string[];
  alias?: { [fromName: string]: string };
  webpack?: Configuration;
  publicAssetURL?: string;
  styleLoaderOptions?: Record<string, unknown>;
  cssLoaderOptions?: Record<string, unknown>;
  forbidEval?: boolean;
  skipBabel?: { package: string; semverRange?: string }[];
  watchDependencies?: (string | string[])[];
+ insertScriptsAt? string;
+ insertStylesAt?: string;
}

```

When unset, we keep today's behavior, which inserts your app scripts after `vendor.js`, app styles after `vendor.css`, test scripts after `test-support.js` and test styles after `test-support.css`.

When set, these are the names of custom elements that appear in your HTML:

```js
autoImport: {
  insertScriptsAt: 'auto-import-scripts',
  insertStylesAt: 'auto-import-stylesheets',
}
```

```html
<auto-import-scripts entrypoint="app"  ></auto-import-scripts>
<auto-import-stylesheets entrypoint="app" ></auto-import-stylesheets>
```

The custom elements have one required attribute: `entrypoint` is the name of a webpack entry point. The default entry points that are always provided by ember-auto-import are named `"app"` and `"tests"`, and they contain everything auto-imported by your app and tests respectively. But you can define other entrypoints under `entry` in your webpack config, and use those names here.

During the build, your custom elements will be replaced with zero or more `<script>` or `<link>` elements, pointing at the built webpack bundles of the given type for the given entry point. Any attributes other than `entryPoint` on the custom element will be copied verbatim onto the resulting `<script>` or `<link>` elements.

This gives you control over both the location and the attributes of the elements ember-auto-import is inserting for you. 

This PR also unlocks a new capability: when you opt into controlling where the entrypoints will get inserted, you can also refer to *custom* entrypoints. For example:

```js
autoImport: {
  insertScriptsAt: 'auto-import-scripts',
  webpack: {
    entry: {
      polyfills: './lib/polyfills.js'
    }
  }
}
```

```js
// lib/polyfills.js
import "core-js/stable";
import "intl";
```

```html
<!-- our special polyfills entrypoint will run before vendor.js, which can be important! -->
<!-- it can also use "nomodule" to not run at all on modern browsers -->
<auto-import-script entrypoint="polyfills" nomodule></auto-import-script>
<script src="{{rootURL}}assets/vendor.js"></script>
<auto-import-script entrypoint="app"></auto-import-script>
<script src="{{rootURL}}assets/app.js"></script>
```

### DIFF
--- a/packages/ember-auto-import/ts/auto-import.ts
+++ b/packages/ember-auto-import/ts/auto-import.ts
@@ -54,7 +54,7 @@ export default class AutoImport implements AutoImportSharedAPI {
     this.packages.add(Package.lookupParentOf(topmostAddon));
     let host = topmostAddon.app;
     this.env = host.env;
-    this.bundles = new BundleConfig(host);
+    this.bundles = new BundleConfig(host.options.outputPaths);
     if (!this.env) {
       throw new Error('Bug in ember-auto-import: did not discover environment');
     }
--- a/packages/ember-auto-import/ts/auto-import.ts
+++ b/packages/ember-auto-import/ts/auto-import.ts
@@ -122,7 +122,7 @@ export default class AutoImport implements AutoImportSharedAPI {
       bundles: this.bundles,
       babelConfig: this.rootPackage.cleanBabelConfig(),
       browserslist: this.rootPackage.browserslist(),
-      publicAssetURL: this.publicAssetURL,
+      publicAssetURL: this.rootPackage.publicAssetURL(),
       webpack,
     });
   }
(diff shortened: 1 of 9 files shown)


ef4:
Next: test end-to-end scenarios including
 - an extra js entrypoint that goes before vendor.js
 - an extra css entrypoint that goes before vendor.css (we should use mini-css-extract-plugin the same way embroider does, and we need to enable it even in dev when using fastboot)

Issue #425, opened 2021-07-20
<my-addon> tried to import "qunit" from addon code, but "qunit" is a devDependency. You may need to move it into dependencies.


NullVoxPopuli:
This is not desired (qunit should remain a dev dep).

But I can't find _where_ this is happening.

Is it possible to log which file the error is about?

<details><summary>Here is the full error</summary>

```

=================================================================================

ENV Summary:

  TIME: Tue Jul 20 2021 18:50:46 GMT-0400 (Eastern Daylight Time)
  TITLE: ember
  ARGV:
  - 🏠/.volta/tools/image/node/14.17.3/bin/node
  - /✂️/{my-monorepo}/shared/packages/{my-addon}/node_modules/.bin/ember
  - build
  EXEC_PATH: 🏠/.volta/tools/image/node/14.17.3/bin/node
  TMPDIR: /tmp
  SHELL: /bin/bash
  PATH:
  - 🏠/.volta/tools/image/yarn/1.22.10/bin
  - 🏠/.volta/tools/image/node/14.17.3/bin
  - 🏠/.volta/bin
  - 🏠/.pyenv/plugins/pyenv-virtualenv/shims
  - 🏠/.pyenv/shims
  - 🏠/.pyenv/bin
  - /✂️/scripts/
  - 🏠/Applications
  - 🏠/apps/phantomjs/bin
  - 🏠/scripts/system-utils
  - 🏠/scripts/git
  - 🏠/scripts/rails
  - 🏠/scripts
  - 🏠/.volta/bin
  - 🏠/.local/bin
  - /usr/local/sbin
  - /usr/local/bin
  - /usr/sbin
  - /usr/bin
  - /sbin
  - /bin
  - /usr/games
  - /usr/local/games
  - /snap/bin
  - 🏠/.fzf/bin
  - /usr/lib/go-1.13/bin
  - /bin
  - 🏠/go/bin
  PLATFORM: linux x64
  FREEMEM: 2951323648
  TOTALMEM: 16790667264
  UPTIME: 64915.81
  LOADAVG: 2.54,1.8,1.54
  CPUS:
  - Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz - 2400
  - Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz - 2400
  - Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz - 2400
[...]

PR #435, opened 2021-08-17
Print file path when an import resolution error occurs

NullVoxPopuli:
Provides insight into issue like what was encountered here: https://github.com/ef4/ember-auto-import/issues/425


ef4:
Instead of adding separate logging, please improve the underlying exceptions to include the `imp.path`:

https://github.com/ef4/ember-auto-import/blob/203ab5886abe93996ed2e2363b0f2d9ee9f8997c/packages/ember-auto-import/ts/package.ts#L281

https://github.com/ef4/ember-auto-import/blob/203ab5886abe93996ed2e2363b0f2d9ee9f8997c/packages/ember-auto-import/ts/package.ts#L301

You'll probably need to add `imp.path` as an additional argument to the `resolve` method on `Package`, but it's only called from two places and both those places have access to `imp.path`.

Issue #270, opened 2020-03-06
Slow warm boot times in large apps


patocallaghan:
Recently we've been some doing some investigating into where we can try and improve our app's warm build speeds. `ember-auto-import` was one candidate addon which appeared to take a substantial amount of time when booting up. Note: _warm_ builds, **NOT** cold or incremental.

We have a large Ember app which has thousands of Javascript and TypeScript files and regularly see ember-auto-import taking a large amount of time at the `analyze` and `append` steps.

I've created a [reproduction repo](https://github.com/patocallaghan/large-app) which essentially contains ~4500 app js files and ~2500 test files. While you can't run this app in the browser this is a somewhat accurate count of the number of JS and test files in our app. Running `yarn ember build` gives a similar build time experience. 

**Repo:** https://github.com/patocallaghan/large-app
**Example BROCCOLI_VIZ example build:** https://github.com/patocallaghan/large-app/blob/master/instrumentation.build.0.json
**Visualization** (from https://heimdalljs.github.io/heimdalljs-visualizer)
![image](https://user-images.githubusercontent.com/685034/76079616-e7fee500-5f9c-11ea-9299-cdbe9fb35390.png)

Running `yarn ember start` (after a cold build)
```
Slowest Nodes (totalTime => 5% )              | Total (avg)         
----------------------------------------------+---------------------
ember-auto-import-analyzer (3)                | 79382ms (26460 ms)  
Bundler (1)                                   | 10364ms             
Babel: large-app (2)                          | 5842ms (2921 ms)  
```

While historically we never thought folks would need to restart their servers often, we've actually found that this is no longer the case. People here will need to restart their servers for various reasons often enough that it has been called out as a productivity killer. Also we run warm builds in our CI setup and any speed improvements there are of great benefit.

Possible suggestions:
* Is it possible to persist caches between warm builds? Killing and restarting your server without making any code changes causes ember-auto-import to redo all of its work.
* Is there a way to parallelize some of the analyze/append work into worker threads? Possibly the analyze AST parsing/updating work?

#### Version
```
yarn ember --version
ember-cli: 3.16.0
node: 12.14.1
os: darwin x64
```

If there are potential changes which could be implemented here I'd be more than happy to offer some time to help out.

ef4:
There is a way to stop paying the parsing cost entirely. We would move
inside the existing ember-cli-babel.

This is pretty straightforward except we need to be really sure to
[...]

Issue #268, opened 2020-02-10
Running out of memory analyzing massive files


EWhite613:
Would like a way to exclude folders (or include).

We have some massive mirage response files in our app. And from the `DEBUG=*` it looks like `ember-auto-import` is trying to analyze those files (resulting in a massive amount of babel logs that look to be parsing).

Followed my an out of memory error.

I know it's `ember-auto-import` because as soon as I remove that addon from the app I can build without issue.

rwjblue:
FWIW, you can't really use `DEBUG=*` due to babel itself emitting so much output (it will often completely crash the process just emitting logging output).

I'd suggest using: `DEBUG=*,-babel`

ef4:
If we're parsing the files, then so is babel and it manages to do it without running out of memory. That makes me wonder if we're retaining something unintentionally for too long.

I'm not sure about an option to ignore directories, because it would result in very confusing conditions when somebody later tries to use an import in an ignored directory and gets no feedback from ember-auto-import about why it won't work.

EWhite613:
@ef4 isn't that just an "Uh duh" moment though?

Like babel also has ignore https://babeljs.io/docs/en/options#ignore I use to just ignore all the mirage files.

There is a warning from babel from all those files but still sound like. they parse it
```
Note: The code generator has deoptimised the styling of <path>/mirage/responses/ws_line_to_line_service-trails.js as it exceeds the max of 500KB
```

ef4:
> I use to just ignore all the mirage files.
> There is a warning from babel

I'm pretty sure you're not successfully causing babel to ignore those files, both because it's still warning and because of how babel runs inside a broccoli pipeline, where the paths it sees are not the ones you authored.

EWhite613:
With the newer auto-import, I didn't have this issue anymore

PR #440, opened 2021-08-20
Optimizing analyzer

ef4:
The plan here is to eliminate the babel parse cost from our analyzer preprocessor. Instead, we will do analysis within the existing ember-cli-babel config, emitting the results in a specially-encoded comment. This ensures that the analysis results have the exact same durability as the babel cache.

We will still have our own followup preprocessor, but it can cheaply gather the comments out of all the files without an expensive parse.


ef4:
Basic idea is now implemented, next we can make the representation more compact and do some benchmarking.

ef4:
I published this as 2.2.0-alpha.0 for benchmarking purposes.

ef4:
Closes #268.
Closes #270.

ef4:
Benchmarks:

App  Description| App Lines of Code | Cold build | Warm build
-----|-----------|-------|----
Small app | 4,600 lines | 10% faster | 20% faster
Medium app | 39,000 lines  | 19% faster | 21% faster

 - Total times from `time ember b`, averaged over three runs
 - Cold builds are with no persistent caches, warm builds are with them
 - "App lines of code" is the total lines of JS or TS in the `/app` folder. Doesn't account for addons or tests, just a rough estimate.

stefanpenner:
I'll review this today.

packages/ember-auto-import/ts/analyzer-syntax.ts
```diff
@@ -0,0 +1,249 @@
...
+  // optional hint for what to name the expression that goes there. It's
+  // optional because in general there may not be an obvious name, but in
+  // practice there often is, and we can aid debuggability by using names that
+  // match the original code.
+  expressionNameHints: (string | null)[];
+}
+
+export type ImportSyntax = LiteralImportSyntax | TemplateImportSyntax;
+
+const marker = 'eaimeta';
```
stefanpenner, line 23: thoughts:

* Do we need a version in the marker? So that upgrades to the format are friendly?
* Do we need a more unique marker? Or do you feel `eaimeta` is enough? my thought was: something like `|ember-auto-import@<meta-version>|`

packages/ember-auto-import/ts/analyzer-syntax.ts
```diff
@@ -0,0 +1,249 @@
...
+    if ('specifier' in imp) {
+      tokens.push(imp.isDynamic ? 1 : 0);
+      tokens.push(imp.specifier);
+    } else {
+      tokens.push(imp.isDynamic ? 3 : 2);
+      tokens.push(imp.cookedQuasis);
+      tokens.push(imp.expressionNameHints);
+    }
+  }
+  return `${marker}${JSON.stringify(tokens).slice(1, -1)}${marker}`;
```
stefanpenner, line 37: How large does this stringified structure get? If it is large, could we gzip it?

packages/ember-auto-import/ts/analyzer-syntax.ts
```diff
@@ -0,0 +1,249 @@
...
+  private reject: (err: any) => void;
+
+  constructor(private source: ReadStream) {
+    let r: (result: ImportSyntax[]) => void, e: (err: any) => void;
+    this.output = new Promise<ImportSyntax[]>((resolve, reject) => {
+      r = resolve;
+      e = reject;
+    });
+    this.resolve = r!;
+    this.reject = e!;
```
stefanpenner, line 89: do we need to observe `stream.finished(stream, options, finishedCallback)` ?

packages/ember-auto-import/ts/analyzer.ts
```diff
@@ -118,36 +73,23 @@ export default class Analyzer extends Plugin {
     return this.modules;
   }
 
-  async build() {
-    await this.setupParser();
-    this.getPatchset().forEach(([operation, relativePath]) => {
-      let outputPath = join(this.outputPath, relativePath);
-
+  async build(...args: unknown[]) {
+    await super.build(...args);
```
stefanpenner, line 77: Whats the TL;DR of using `Funnel` as the base class?  It sorta hurts my brain to think about how that works. It seems like we would end up potentially end up double walking the input right?

stefanpenner (changes requested):

packages/ember-auto-import/ts/analyzer.ts
ef4: Funnel with no options is optimized to be the fastest pass through it can, including symlinking the whole tree when it can.

packages/ember-auto-import/ts/analyzer-syntax.ts
ef4: It's smaller than the import statements it is summarizing, so I think not worth zipping.

packages/ember-auto-import/ts/analyzer.ts
ef4: It would be nice if broccoli had a first-class read-only primitive but I don't think it does. So a read-only transform like ours is obligated to propagate changes forward and needs to do so efficiently.

packages/ember-auto-import/ts/analyzer-syntax.ts
ef4: This method only accepts fs.ReadStream, not a generic readable. ReadStream always emits close.

packages/ember-auto-import/ts/analyzer-syntax.ts
stefanpenner: @ef4 awsome!

packages/ember-auto-import/ts/analyzer.ts
stefanpenner: @ef4 oh I get it thanks. That is a good "hack",  and makes sense in this context.

stefanpenner:
I'm glad to see this has a nice performance boost.

packages/ember-auto-import/ts/analyzer-syntax.ts
ef4: Versioning is a good question. 

A single copy of ember-auto-import does both the reading and writing, so that is not an issue. But since the babel cache is persistent, you could find things in the cache that were written by a different version. I think the best way to address that is to make sure our babel plugin config itself changes if the protocol changes. We can include the protocol version in the babel plugin config and that should take care of it, because AFAIK changing the babel config should invalidate the babel persistent cache.

Regarding uniqueness: because we prepend at the very beginning of each module and our reading side consumes the first copy found, we only need to resist collisions with anything that gets prepended in front of us. But since that includes the AMD wrapper and that includes everybody's module names, it's definitely possible to put stuff there that would collide.

I'll tackle both of these by picking a weird marker and by making the marker an explicitly piece of the babel config.

ef4:
As a followup for people following this issue: in https://github.com/ef4/ember-auto-import/pull/442 I had to re-add the more expensive parsing when we encounter addons with older copies of ember-auto-import. This means you will get a more modest performance improvement if you only upgrade your app while your addons are on an older version of ember-auto-import.

boris-petrov:
Is it possible to backport this to the [v1.x](https://github.com/ef4/ember-auto-import/tree/v1.x) branch? Some of us still can't update and we would like to benefit from this. :)

ef4:
Yes, I would take a PR back-porting this to the v1.x branch.

PR #442, opened 2021-08-31
mixing of old and new analyzer patterns

ef4:
The changes in #440 can break when an addon is using a 2.x copy of ember-auto-import that predates the changes. The problem is that older copies will run their own `included` hook which does not install the babel plugin.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30834,6 +30834,195 @@
         "node": ">=0.10.0"
       }
     },
+    "node_modules/old-analyzer": {
+      "name": "ember-auto-import",
+      "version": "2.1.0",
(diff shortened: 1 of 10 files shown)


ef4:
Even if we ensure that the babel plugin is installed, existing versions of ember-auto-import are configured to run the analyzer *before* babel, whereas our new technique requires that we run *after*. 

So I don't think we can get a seamless switch to the new faster analyzer. Instead, when we find that the analyzer is running with no babel plugin before it we will need to continue doing the expensive parse.  People can avoid that cost by taking the semver-compatible upgrade, which should be as simple as re-rolling their lockfile.

PR #451, opened 2021-10-01
Support .mjs extensions

NullVoxPopuli:
If I'm reading this right, https://webpack.js.org/guides/ecma-script-modules/

*.mjs is supported (as modules) no matter what?

### DIFF
--- a/packages/ember-auto-import/ts/webpack.ts
+++ b/packages/ember-auto-import/ts/webpack.ts
@@ -160,7 +160,7 @@ export default class WebpackBundler extends Plugin implements Bundler {
         },
       },
       resolve: {
-        extensions: ['.js', '.ts', '.json'],
+        extensions: ['.js', '.mjs', '.ts', '.json'],
         mainFields: ['browser', 'module', 'main'],
         alias: Object.assign({}, ...[...this.opts.packages].map(pkg => pkg.aliases).filter(Boolean)),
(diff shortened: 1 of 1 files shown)


ef4:
Those docs are talking about controlling how webpack interprets files as either modules or commonjs. It always interprets mjs as modules. But that says nothing about whether they will be resolved.

For the `resolve.extensions`, our default value matches the webpack defaults, which do not include `.mjs`.

If you encounter a third-party dependency that needs mjs, you should be able to override the setting?

```js
  autoImport: {
    webpack: {
      resolve: {
        extensions: ['.js', '.mjs', '.json', '.wasm'],
      }
    }
  }
```

NullVoxPopuli:
would it be harmful to support by default though?

ef4:
> would it be harmful to support by default though?

Yes, it probably would. This is a setting with global ramifications and this is likely a breaking change because people could have inert `.mjs` files sitting in their dependencies right now that would start taking precedence instead.

I don't think there's one clear "best" choice for this setting, which is why I consider sticking to the webpack default the safest choice. Libraries that break under the webpack defaults are more likely to get fixed because more people are testing them under those conditions. It's a safety-of-the-herd thing.

So my suggestion is to PR this change to webpack itself. We will follow the webpack default if it changes.

Issue #449, opened 2021-09-21
Analyzer messing up import names at the chunk boundaries


vlascik:
I'm trying to import and customize https://github.com/formula/formula, a kind of an formula language, similar to Excel's formulas, and to do that, I have a fairly large file reexporting the list of the included functions (subset of src/functions.js).

However, there is a bug in EAI `analyzer-syntax.ts`, where `matchesAtEnd()` returns `1` if chunk ends in partial marker's first letter. The problem is, when chunk ends with e.g. the same letter (`e`) as the beginning of the marker, analyzer is then going to cut off the last character of the chunk. Which is then missing after the list of imports gets reassembled from the state:

```
state.meta = Array(7)
...
2 = e",0,"formula/src/isdate",0,"formula/src/isemail",0,"formula/src/isemail",0,"formula/src/isemail",0,"formula/src/isempty",0,"formula/src/isempty",0,"formula/src/isempty",0,"formula/src/iserror",0,"formula/src/iserror",0,"formula/src/iserror",0,"formula/src/iseven",0,"formula/src/iseven",0,"formula/src/iseven",0,"formula/src/isfalsy",0,"formula/src/isfalsy",0,"formula/src/isfalsy",0,"formula/src/isfunction",0,"formula/src/isfunction",0,"formula/src/isfunction",0,"formula/src/isleapyear",0,"formula/src/isleapyear",0,"formula/src/isleapyear",0,"formula/src/islowercase",0,"formula/src/islowercase",0,"formula/src/islowercase",0,"formula/src/isobject",0,"formula/src/isobject",0,"formula/src/isna",0,"formula/src/isna",0,"formula/src/isna",0,"formula/src/isnan",0,"formula/src/isnan",0,"formula/src/isnan",0,"formula/src/isnumber",0,"formula/src/isnumber",0,"formula/src/isnumber",0,"formula/src/isodd",0,"formula/src/isodd",0,"formula/src/isodd",0,"formula/src/isoweeknum",0,"formula/src/isoweeknum",0,"formula/src/isowe (length: 1023)
```
```
[...]

PR #454, opened 2021-10-04
Prevent loss of characters in case of false end-partial-match

timmorey:
Attempted fix for https://github.com/ef4/ember-auto-import/issues/449

I believe I ran into the same issue, and built on the debugging that @vlascik had done.  I don't have a deep understanding of how this code fits into the overall build pipeline, but was able to wrap my head around the code in this module enough to propose a fix.

This addresses the case where we found a partial match for `MARKER` at the end of the last chunk, but now that we're looking in the next chunk we've realized it wasn't really a match.  The fix is to make sure the characters we sliced off of the last chunk don't get lost.

### DIFF
--- a/packages/ember-auto-import/ts/analyzer-syntax.ts
+++ b/packages/ember-auto-import/ts/analyzer-syntax.ts
@@ -181,10 +181,11 @@ class Deserializer {
             meta: state.meta.join(''),
           };
         } else {
-          // partial match failed to complete
+          // partial match failed to complete, so we need to replace the partial
+          // marker match we stripped off the last chunk
           this.state = {
             type: 'finding-end',
-            meta: state.meta,
+            meta: [...state.meta, MARKER.slice(0, state.partialMatch)],
           };
           return this.consumeChunk(chunk);
         }


ef4:
Thanks! Nice work.

The tests for this area are here: https://github.com/ef4/ember-auto-import/blob/a88004f656df9e08d3d5e7e3eb337e0cd5727b9e/packages/ember-auto-import/ts/tests/analyzer-test.ts#L379-L490

If you can get a test case in here that fails without your fix and passes with it, we'll be good to go.

timmorey:
@ef4 thanks for the pointer.  Test added.

packages/ember-auto-import/ts/tests/analyzer-test.ts
```diff
@@ -487,4 +487,17 @@ Qmodule('analyzer-deserialize', function () {
     let result = await deserialize(source([`stuff stuff stuff ${meta.slice(0, -2)}`, meta.slice(-2)]));
     assert.deepEqual(result, sampleData());
   });
+
+  test('false end marker at end of chunk', async function (assert) {
+    const meta = serialize(sampleData());
+    assert.ok(
+      meta.slice(MARKER.length, -MARKER.length).indexOf(MARKER[0]) > -1,
+      'serialized sample data must contain first character of MARKER somewhere between boundary markers for test to have meaning'
+    );
```
timmorey, line 496: This assertion isn't really checking the result of running code -- its just checking that our sample data is sufficient for this test case.  If `MARKER` or the sample data were changed in the future, such that the first character of `MARKER` didn't show up somewhere in the serialized data, this test would become invalid.

PR #458, opened 2021-10-11
Fix not passing the correct options to `style-loader` and `css-loader`

### LINKED ISSUE #416: Cannot use `styleLoaderOptions`
Thanks for releasing a new version with the [commit](https://github.com/ef4/ember-auto-import/commit/2bf7100c954316435f19c7e4535ab5db9b376429) for passing options to the built-in `style-loader`!

However, it doesn't work. Setting (even an empty) `styleLoaderOptions` key in `autoImport` leads to:

```
ValidationError: Invalid options object. Style Loader has been initialized using an options object that does not match the API schema.
 - options has an unknown property '_babelMajorVersion'. These properties are valid:
   object { injectType?, attributes?, insert?, base?, esModule?, modules? }
```

I'm not sure where this `_babelMajorVersion` comes from... @ef4 - any ideas?

boris-petrov:
Fixes #416 

cc @ef4


boris-petrov:
Thank you!

Issue #462, opened 2021-10-20
Build fails on windows


timmorey:
We have some users that build our app on windows, and builds have started to fail since we upgraded to ember-auto-import 2.x.

This appears to be caused by the line https://github.com/ef4/ember-auto-import/blob/497d5c56e62dc780d3ccd04e9532872bc7332890/packages/ember-auto-import/ts/webpack.ts#L366, which attempts to create a directory symlink.  This requires admin privileges in windows.

We get errors that look like: `EPERM: operation not permitted, symlink 'C:\Users\timo7353\ember-auto-import-on-windows\node_modules\qunit' -> 'C:\Users\timo7353\AppData\Local\Temp\broccoli-67763erePnZtVDr0\cache-307-webpack_bundler_ember_auto_import_webpack\node_modules\qunit'`

I reproduced in a simple app: https://github.com/timmorey/ember-auto-import-on-windows.  All I did was create a new ember 3.28 app and then pull in the latest ember-auto-import/webpack.  Running `ember test` on a windows 10 box throws the error shown above.

ef4:
While I do consider this a real bug and we can fix it, you should know that if you're running ember-cli on windows and haven't enabled the ability to create symlinks your build performance may be relatively horrible. See https://cli.emberjs.com/release/appendix/windows/#whatcausesthebuildslowdown

timmorey:
Thanks for the tip about ember-cli+windows.  None of us use windows for development.  There is another team that builds our app for some deployments, and they use windows for just about everything.  I'll send that link to them.

I took [your changes](https://github.com/ef4/ember-auto-import/pull/463) for a test run, and it looks like it fixes the error we were seeing.

charlesfries:
@timmorey any tips on installing `ember-auto-import` at this commit? I am also experiencing this symlink issue on Windows and I can't figure out how to install EAI from URL since it's a monorepo.

PR #463, opened 2021-10-20
windows symlinks->junctions

ef4:
This may be a fix for #462.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40447,7 +40447,7 @@
       }
     },
     "packages/ember-auto-import": {
-      "version": "2.2.2",
+      "version": "2.2.3",
       "license": "MIT",
       "dependencies": {
         "@babel/core": "^7.1.6",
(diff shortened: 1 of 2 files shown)


charlesfries:
@timmorey any tips on installing `ember-auto-import` at this commit? I am also experiencing this symlink issue on Windows and I can't figure out how to install EAI from URL since it's a monorepo.

timmorey:
@charlesfries unfortunately, I haven't had success doing that either.  I even attempted to make a branch in my fork that had all of the built assets and the structure that was expected for the package, but then I started running into build errors in https://github.com/ef4/ember-auto-import/blob/main/packages/ember-auto-import/ts/leader.ts since I was referencing a github branch in my package.json instead of a semver range.

When I validated these changes in my build, I did so by manually editing files in my `node_modules` directory to update the affected code - not a viable production workflow.

timmorey:
@ef4 I looked into the reported test failure, and it appears unrelated to the changes.  It looks like there was a timeout when installing dependencies: https://github.com/ef4/ember-auto-import/runs/3956772037?check_suite_focus=true#step:5:32. Re-running the test may be sufficient.

ef4:
Thanks, this is released in 2.2.4.

PR #466, opened 2021-10-26
Avoid EBADF on ReadStream early close in Node 12

maxfierke:
On Node <= 14, there seems to be a potential for EBADF to get raised
after a ReadStream has been closed (https://github.com/nodejs/node/issues/25335)

We can workaround this causing problems by finishing the deserializer's operations
rather than calling 'destroy' on the ReadStream mid-read. This way the Promise gets
resolved with the result before a potential EBADF can occur.

Then we can simply destroy the ReadStream explicitly when we know we're done with it.

Also made sure to add `emitClose` for Node 12, which is `false` by default until Node 14 and might be causing the close event handlers not to get processed. (Though TypeScript seemed upset about adding that, despite it being a documented option for `createReadStream`, so had to slap on a `// @ts-ignore`. Let me know if there's a way to fix that.)

Seems to fix #464, as I haven't been able to reproduce the EBADF after re-running a couple dozen times (where previously I would get it 25-50% of the time)

PR #470, opened 2021-11-14
Prevent ebadf

ef4:
...by always resolving our promise before destroying the source.

This incorporates the insight from https://github.com/ef4/ember-auto-import/pull/466 but in a style that keeps all our state together in `this.state`.


ef4:
(This PR was just to get CI testing this example, merging onto #466 to get it landed.)

PR #473, opened 2021-12-07
chore(documentation): fix upgrade guide link

lifeart:


ef4:
This link already works when used from the main repo page: https://github.com/ef4/ember-auto-import

I think this would break it there?

I don't love the symlink strategy here, open to other suggestions.

PR #489, opened 2022-01-19
Update to latest @embroider/internals and @embroider/macros

SergeAstapov:
A follow up to https://github.com/embroider-build/embroider/releases/tag/v1.0.0


jrjohnson:
Would be awesome if this can be back ported to `1.12.x` as well to cover addons getting it through the `ember-cli@3.28` blueprint. Would it be helpful for me to open a separate PR to do that?

ef4:
We can do the backport too but it's not as simple. The APIs that were being used out of `@embroider/core` have since moved into `@embroider/shared-internals` and they changed slightly.

ef4:
1.12.1 is released with the embroider 1.0 backport.

PR #496, opened 2022-01-25
don't let broccoli-asset-rev mess with css chunks

### LINKED ISSUE #493: [2.4.0] Building in `production` environment fingerprints CSS chunk files
This was not true before. Starting with 2.4.0, building in `production` environment leads to files like this:

```
chunk.4949.57562f2dfce27933423a-074361039f03e2057723a63fa3016515.css
```

The `-074361039f03e2057723a63fa3016515` part is new (and shouldn't be there). This means that this chunk was passed to `ember-cli`'s fingerprinter which should not happen.

Doesn't happen with JS files, only with CSS ones.

ef4:
We already have code to defend our JS chunks from getting renamed by broccoli-asset-rev, but we need the same for any emitted CSS chunks.

Fixes #493.

### DIFF
--- a/packages/ember-auto-import/ts/auto-import.ts
+++ b/packages/ember-auto-import/ts/auto-import.ts
@@ -208,7 +208,7 @@ export default class AutoImport implements AutoImportSharedAPI {
   // have their own webpack-generated hashes and (2) the runtime loader code
   // can't easily be told about broccoli-asset-rev's hashes.
   private configureFingerprints(host: AppInstance) {
-    let patterns = ['assets/chunk.*.js', 'assets/chunk..*.css'];
+    let patterns = ['assets/chunk.*.js', 'assets/chunk.*.css'];
     if (!host.options.fingerprint) {
       host.options.fingerprint = {};
     }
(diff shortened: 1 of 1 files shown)


packages/ember-auto-import/ts/auto-import.ts
```diff
@@ -208,14 +208,16 @@ export default class AutoImport implements AutoImportSharedAPI {
   // have their own webpack-generated hashes and (2) the runtime loader code
   // can't easily be told about broccoli-asset-rev's hashes.
   private configureFingerprints(host: AppInstance) {
-    let pattern = 'assets/chunk.*.js';
+    let patterns = ['assets/chunk.*.js', 'assets/chunk..*.css'];
```
Windvis, line 211:
```suggestion
    let patterns = ['assets/chunk.*.js', 'assets/chunk.*.css'];
```
I think the extra `.` isn't needed here?

SergeAstapov: as chunk files named as `chunk.[some hash here].css` seem actually legit? 🤔

Windvis: We were running into this issue in one of our apps so I tried adding the pattern from this PR in our own config but it didn't work. Removing the double `.` next to the `*` made it work as expected.

The chunks look like this `/assets/chunk.536.b58262bb39d5d25c6454.css` so I think it either needs to be the pattern I suggested (which removes the double `.`) or something like `assets/chunk.*.*.css` which is a more strict match but I would go for consistency with the js pattern anyways.

ef4:
Yup, that extra dot was just a typo.

PR #502, opened 2022-02-04
Bump `fs-extra` to v10 and `resolve-package-path` to v4

SergeAstapov:
Not sure if this was intentionally omit.

This would help with dependency tree deduce


ef4:
Thanks.

PR #507, opened 2022-03-11
Clarify upgrade guide for addon

ctjhoa:
Clarify the fact that, if you only use `ember-auto-import` as `devDependency` you are not forced to do a major release as a maintainer.


ef4:
Thanks, this makes sense.

PR #509, opened 2022-03-18
Introduce option to allow skipping insertion of styles and scripts

dgeb:
This PR introduces a new ability to entirely skip insertion of scripts and/or styles by the `Inserter`. 

The use case I am targeting is a static host site which is developed using fastboot and published with prember. Ember is not used for dynamic interactivity, which is provided using glimmerx components in this case (although these components could just as easily be written with another framework and layered onto the static pages).

I have attempted to make supporting this use case as lightweight as possible by re-using existing options, `insertScriptsAt` and `insertStylesAt`, and allowing a new possible value for them: `null`. This value indicates to the `Inserter` that insertion of scripts / styles should be skipped when processing html.

I have marked this as a draft because I have not yet added tests or a description to the README, but would be glad to do so if this looks like an acceptable approach. If there's a better approach you'd like me to take to achieve a similar goal, please let me know!

### DIFF
--- a/packages/ember-auto-import/ts/inserter.ts
+++ b/packages/ember-auto-import/ts/inserter.ts
@@ -12,8 +12,8 @@ const debug = makeDebug('ember-auto-import:inserter');
 
 export interface InserterOptions {
   publicAssetURL: string;
-  insertScriptsAt: string | undefined;
-  insertStylesAt: string | undefined;
+  insertScriptsAt: string | null | undefined;
+  insertStylesAt: string | null | undefined;
 }
 
 interface ScriptTarget {
@@ -98,7 +98,7 @@ export class Inserter extends Plugin {
         `looking for custom script element: %s`,
         this.options.insertScriptsAt
       );
-    } else {
+    } else if (this.options.insertScriptsAt === undefined) {
       debug(
         `looking for scripts with src: %s`,
         targets.scripts.map((s) => s.afterFile).filter(Boolean)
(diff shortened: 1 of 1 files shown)


ef4:
@dgeb and I discussed this over discord, summarizing as: both embroider and upcoming versions of ember-cli-fastboot are likely to not work if you don't have script tags in the HTML, because they treat the HTML as the source of truth.

Probably the best path forward involves post-processing the build output after ember-auto-import is done with it.

Fastboot supports `<fastboot-script>` in the HTML build output as a way to tell it to run a script that should be stripped out before going to the browser.

dgeb:
@ef4 Thanks for the discussion and recommendations.

> Fastboot supports `<fastboot-script>` in the HTML build output as a way to tell it to run a script that should be stripped out before going to the browser.

I'd love to be able to take this approach. It seems ideal to allow `ember-auto-import` to check for app scripts in either `script` or `fastboot-script` elements. However, if app scripts are only in `fastboot-scripts`, the `Inserter` does not identify them and will throw an error (`ember-auto-import could not find a place to insert app scripts in ${filename}`).

> Probably the best path forward involves post-processing the build output after ember-auto-import is done with it.

I've taken this approach and gotten things working with a simple post-processing step that strips out all lines with `script` elements that reference `src="/assets/`

```js
// ember-cli-build.js
'use strict';

const BroccoliReplace = require('broccoli-replace');
const EmberApp = require('ember-cli/lib/broccoli/ember-app');

module.exports = function (defaults) {
  const app = new EmberApp(defaults, {});

  return new BroccoliReplace(app.toTree(), {
    files: ['index.html'],
    patterns: [
      {
        match: /^.*\bscript src="\/assets\/\b.*$/gm,
        replacement: ''
      }
    ]
  });
};
```

If I can improve and generalize this approach, I will document it further.

Issue #469, opened 2021-11-09
Editing CSS files causes whole page to reload


x8BitRain:
Edit:

Creating a new project and adding the latest ember-auto-import and webpack 5 causes full page reload when editing CSS instead of just hot reloading the CSS without refreshing the page. Is there a way to avoid this? Thanks.

Package.json:

    {
      "name": "thing",
      "version": "0.0.0",
      "private": true,
      "description": "Small description for thing goes here",
      "repository": "",
      "license": "MIT",
      "author": "",
      "directories": {
        "doc": "doc",
        "test": "tests"
      },
      "scripts": {
        "build": "ember build --environment=production",
        "lint": "npm-run-all --aggregate-output --continue-on-error --parallel 'lint:!(fix)'",
        "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*:fix",
        "lint:hbs": "ember-template-lint .",
        "lint:hbs:fix": "ember-template-lint . --fix",
        "lint:js": "eslint . --cache",
        "lint:js:fix": "eslint . --fix",
        "start": "ember serve",
        "test": "npm-run-all lint test:*",
        "test:ember": "ember test"
      },
      "devDependencies": {
        "@ember/optional-features": "^2.0.0",
        "@ember/test-helpers": "^2.2.0",
        "@glimmer/component": "^1.0.3",
        "@glimmer/tracking": "^1.0.3",
        "babel-eslint": "^10.1.0",
        "broccoli-asset-rev": "^3.0.0",
        "ember-auto-import": "^2.0.0",
        "ember-cli": "~3.25.1",
        "ember-cli-app-version": "^4.0.0",
        "ember-cli-babel": "^7.23.1",
        "ember-cli-dependency-checker": "^3.2.0",
        "ember-cli-htmlbars": "^5.3.2",
        "ember-cli-inject-live-reload": "^2.0.2",
        "ember-cli-sri": "^2.1.1",
        "ember-cli-terser": "^4.0.1",
        "ember-data": "~3.25.0",
        "ember-export-application-global": "^2.0.1",
        "ember-fetch": "^8.0.4",
        "ember-load-initializers": "^2.1.2",
        "ember-maybe-import-regenerator": "^0.1.6",
        "ember-page-title": "^6.2.1",
        "ember-qunit": "^5.1.2",
        "ember-resolver": "^8.0.2",
        "ember-source": "~3.25.1",
        "ember-template-lint": "^2.18.1",
[...]

PR #510, opened 2022-03-22
avoid unnecssary full page reloads

ef4:
Possible fix for #469

### DIFF
--- a/packages/ember-auto-import/ts/inserter.ts
+++ b/packages/ember-auto-import/ts/inserter.ts
@@ -2,7 +2,7 @@ import type { InputNode } from 'broccoli-node-api';
 import Plugin from 'broccoli-plugin';
 import makeDebug from 'debug';
 import { existsSync, readFileSync } from 'fs';
-import { outputFileSync, readJSONSync, writeJSONSync } from 'fs-extra';
+import { outputFileSync, readJSONSync } from 'fs-extra';
 import { join } from 'path';
 import parse5 from 'parse5';
 import BundleConfig from './bundle-config';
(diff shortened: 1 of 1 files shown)


gilest:
Looks good. Tried to test with the repro in #469 but ran into roadblocks installing `ember-auto-import` from git branch (monorepo package manager issues) or local directory (breaks version check).

Let me know if there's a way I can test or verify.

SergeAstapov:
> Looks good. Tried to test with the repro in #469 but ran into roadblocks installing `ember-auto-import` from git branch (monorepo issue) or local directory (breaks version check).

just an idea, you may try to run `npm pack` in ember-auto-import branch with fix to get tgz file and use `file:` scheme in package.json (followed by yarn i/npm i/pnpm i) to point to that file from the app to test the changes.
Otherwise there is no easy way to point to package from monorepo (git submodules may be other option but that’s way to wild 😝 imo)

gilest:
> you may try to run `npm pack` in ember-auto-import branch with the fix to get tgz file and use `file:` scheme in package.json (followed by yarn i/npm i/pnpm i) to point to that file from the app to test the changes

Ah – thanks! I hadn't tried this... but it still fails the version check.

> ember-auto-import was unable to find any copy of itself that satisfies the app's requested range of file:/Users/giles/code/ember-auto-import-2.4.0.tgz

Anyway I patched past the version check by editing the `leader.js` in `node_modules` and hardcoding the `appCandidate` version.

Fix works as expected 🙌🏻

ef4:
This is released now anyway. Thanks.

Issue #503, opened 2022-02-12
Let webpack handle test vs app layering


ef4:
At present, we generate two separate entrypoints to hand to webpack (one for app, one for tests). We follow the same pattern that normal ember apps do, such that the tests are allowed to assume that the app is necessarily always loaded first, so a dependency that is used by the tests should be left out of the test entrypoint if it's also used by the app.

However, it would be better to switch to layering the tests and app the way embroider does, by creating an explicit dependency from the test entrypoint to the app entrypoint. This lets webpack handle the deduplication, and it can take into account more resolution detail than we can. 

A case that fails today is having app code that imports from "some-dependency/thing.js" and test code that imports from "some-dependency/thing". Both resolve to the same module, but it will actually load and execute twice because ember-auto-import fails to deduplicate it.

ef4:
@SergeAstapov pointed out that his causes even more significant trouble because having separate entrypoints for app and tests on the same page causes webpack to have duplicated runtime caching, so even when modules aren't duplicated they can get evaluated multiple times and not share state.

knownasilya:
Is this why I'm seeing errors like this: https://github.com/knownasilya/ember-toggle/runs/8071997603?check_suite_focus=true#step:6:74

ef4:
No, I think that is probably https://github.com/ef4/ember-auto-import/issues/504

runspired:
even if we solve why ember-auto-import doesn't work when treeForAddon is custom (all babel plugins present, but I think the other custom tree stuff auto-import does is the issue, see #543) this issue will prevent EmberData from converting anything to V2 Addons as we rely on module state for quite a large number of things.

PR #512, opened 2022-03-29
Restructure entry file bundle logic to support "layering"

thoov:
Currently, ember-auto-import has two conceptual buckets where it places the imports it has found (an **app** bucket and a **tests** bucket). While this works, it has two potential problems. The first is with imports from within lazy engines as they get placed inside of the "app" bucket. This means that instead of these asset loading lazily they are instead eagerly loaded as part of the main app bundle. The second problem is that modules can be "duplicated" in each bucket and can cause problems with module states' being different depending on the importer.

In order to fix this, this PR creates an individual entry file (or bucket) for every addon that is doing an import. It then creates import links from each entry file to its rightful parent (ie a nested addon from a lazy engine would only be imported from the lazy engine's entry file and then the lazy engine's entry file is import from the app's entry file dynamically). Additionally, the tests entry file now imports the app entry file which creates a conceptual graph for webpack to properly traverse such that modules will not be duplicated.

Fixes: #503

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5125,6 +5125,12 @@
         "babel-runtime": "^6.22.0"
       }
     },
+    "node_modules/babel-plugin-compact-reexports": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/babel-plugin-compact-reexports/-/babel-plugin-compact-reexports-1.1.0.tgz",
+      "integrity": "sha512-+KgjNJ5yMeZzJxYZdLEy9m82m92aL7FLvNJcK6dYJbW06t+UTpFJ2FVSs35zMfURcPnrQELYhLG4VC+kt/4gvw==",
+      "dev": true
+    },
     "node_modules/babel-plugin-debug-macros": {
       "version": "0.3.4",
       "resolved": "https://registry.npmjs.org/babel-plugin-debug-macros/-/babel-plugin-debug-macros-0.3.4.tgz",
@@ -6417,6 +6423,68 @@
         "object-assign": "^4.1.0",
         "path-posix": "^1.0.0",
         "symlink-or-copy": "^1.1.8"
+      }
+    },
+    "node_modules/broccoli-dependency-funnel": {
+      "version": "2.1.2",
+      "resolved": "https://registry.npmjs.org/broccoli-dependency-funnel/-/broccoli-dependency-funnel-2.1.2.tgz",
+      "integrity": "sha512-k6b0OnNuRcUnJ9TXA0o6RvqXOkTQ6APKoLsZeMJHAe/YjLjE1uTlfw4Z88GfGmi8gwtLHdnkrhBoJ7YdIkcVZA==",
+      "dev": true,
+      "dependencies": {
+        "broccoli-plugin": "^1.3.1",
+        "fs-tree-diff": "^0.5.9",
+        "heimdalljs": "^0.2.5",
+        "heimdalljs-logger": "^0.1.9",
+        "mkdirp": "^0.5.1",
+        "mr-dep-walk": "^1.4.0",
+        "path-posix": "^1.0.0",
+        "rimraf": "^2.6.2",
+        "symlink-or-copy": "^1.2.0"
+      },
+      "engines": {
+        "node": ">= 6.0.0"
+      }
+    },
+    "node_modules/broccoli-dependency-funnel/node_modules/broccoli-plugin": {
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/broccoli-plugin/-/broccoli-plugin-1.3.1.tgz",
+      "integrity": "sha512-DW8XASZkmorp+q7J4EeDEZz+LoyKLAd2XZULXyD9l4m9/hAKV3vjHmB1kiUshcWAYMgTP1m2i4NnqCE/23h6AQ==",
+      "dev": true,
+      "dependencies": {
+        "promise-map-series": "^0.2.1",
+        "quick-temp": "^0.1.3",
+        "rimraf": "^2.3.4",
+        "symlink-or-copy": "^1.1.8"
+      }
+    },
+    "node_modules/broccoli-dependency-funnel/node_modules/fs-tree-diff": {
+      "version": "0.5.9",
+      "resolved": "https://registry.npmjs.org/fs-tree-diff/-/fs-tree-diff-0.5.9.tgz",
+      "integrity": "sha512-872G8ax0kHh01m9n/2KDzgYwouKza0Ad9iFltBpNykvROvf2AGtoOzPJgGx125aolGPER3JuC7uZFrQ7bG1AZw==",
+      "dev": true,
+      "dependencies": {
+        "heimdalljs-logger": "^0.1.7",
+        "object-assign": "^4.1.0",
+        "path-posix": "^1.0.0",
+        "symlink-or-copy": "^1.1.8"
+      }
+    },
+    "node_modules/broccoli-dependency-funnel/node_modules/minimist": {
+      "version": "1.2.6",
+      "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.6.tgz",
+      "integrity": "sha512-Jsjnk4bw3YJqYzbdyBiNsPWHPfO++UGG749Cxs6peCu5Xg4nrena6OVxOYxrQTqww0Jmwt+Ref8rggumkTLz9Q==",
+      "dev": true
+    },
+    "node_modules/broccoli-dependency-funnel/node_modules/mkdirp": {
+      "version": "0.5.6",
+      "resolved": "https://registry.npmjs.org/mkdirp/-/mkdirp-0.5.6.tgz",
+      "integrity": "sha512-FP+p8RB8OWpF3YZBCrP5gtADmtXApB5AMLn+vdyA+PyxCjrCs00mjyUozssO33cwDeT3wNGdLxJ5M//YqtHAJw==",
+      "dev": true,
+      "dependencies": {
+        "minimist": "^1.2.6"
+      },
+      "bin": {
+        "mkdirp": "bin/cmd.js"
       }
     },
     "node_modules/broccoli-file-creator": {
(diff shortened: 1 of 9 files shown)


thoov:
@ef4: When you have a chance could you review this?

package-lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "ember-auto-import",
+  "name": "@thoov/ember-auto-import",
```
rwjblue, line 2: Probably should be reverted?

packages/ember-auto-import/ts/splitter.ts
```diff
@@ -252,13 +252,36 @@ export default class Splitter {
...
+        let bundleNames = this.chooseBundle(staticUses);
+        bundleNames.forEach((name) => {
+          if (!deps.has(name)) {
+            deps.set(name, {
+              staticImports: [],
+              staticTemplateImports: [],
+              dynamicImports: [],
+              dynamicTemplateImports: [],
+            });
+          }
```
rwjblue, line 266: Maybe extract this to a smaller helper function?

I'm thinking it might look like:

```js
bundleNames.forEach((name) => {
  const dep = ensureDep(deps, name);
  dep.staticImports.push(target); 
});
```

package-lock.json
thoov: Thanks, fixed

packages/ember-auto-import/ts/splitter.ts
thoov: Added this helper function

packages/ember-auto-import/ts/auto-import.ts
```diff
@@ -66,7 +66,16 @@ export default class AutoImport implements AutoImportSharedAPI {
     this.packages.add(Package.lookupParentOf(topmostAddon));
     let host = topmostAddon.app;
     this.env = host.env;
-    this.bundles = new BundleConfig(host.options.outputPaths);
+    let lazyEngines = host.project.addons.filter(
+      (addon: any) =>
+        addon.options &&
+        addon.options.lazyLoading &&
+        addon.options.lazyLoading.enabled
+    );
```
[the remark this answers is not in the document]
thoov: Changed this to be a recursive lookup

ef4:
@thoov the above commit adds the failure case I mentioned. It does indeed break.

thoov:
@ef4 could you review this again. I got the test case you added to pass using [`dependsOn`](https://webpack.js.org/configuration/entry-context/#dependencies).

packages/ember-auto-import/ts/splitter.ts
```diff
@@ -315,6 +365,10 @@ export default class Splitter {
         ? this.options.bundles.bundleForPath(usage.path)
         : this.options.bundles.bundleForTreeType(usage.treeType);
 
+    if (usage.package.isLazyEngine) {
```
ef4, line 368: I don't think this is correct. It only handles the case where the lazy engine itself uses ember-auto-import. But the lazy engine can use an addon that uses an addon that uses ember-auto-import. Should the lazy engine's bundle get those deps? Depends on who else consumes that chain of addons.

This is why I suggested making every package, not just lazy engines, emit their own corresponding module for webpack to process. It's much easier to reason about, because all our decision making is local-only, and the global reasoning can be left to webpack to optimize.

packages/ember-auto-import/ts/webpack.ts
```diff
@@ -89,7 +95,7 @@ window._eai_d = define;
 export default class WebpackBundler extends Plugin implements Bundler {
   private state:
     | {
-        webpack: Compiler;
+        webpack: MultiCompiler;
```
ef4, line 98: I think MultiCompiler means we're running the whole webpack build separately multiple times? That should not be necessary. We can keep adding more entrypoints to the same build instead.

packages/ember-auto-import/ts/webpack.ts
```diff
@@ -129,18 +135,59 @@ export default class WebpackBundler extends Plugin implements Bundler {
...
+        };
 
-    let entry: { [name: string]: string[] } = {};
-    this.opts.bundles.names.forEach((bundle) => {
-      entry[bundle] = [
-        join(stagingDir, 'l.js'),
-        join(stagingDir, `${bundle}.js`),
-      ];
+        if (bundle !== 'app') {
+          // TODO: should this apply to custom entry props?
```
ef4, line 159: Probably no, custom entrypoints mostly want to be entirely independent.

packages/ember-auto-import/ts/webpack.ts
```diff
@@ -129,18 +135,59 @@ export default class WebpackBundler extends Plugin implements Bundler {
...
+          entry[bundle].dependOn = ['app'];
```
ef4, line 160: I think a lazy engine can depend on a lazy engine too. The dependency graph is more complicated than just app vs flat set of lazy engines.

test-scenarios/engines-test.ts
```diff
@@ -0,0 +1,243 @@
...
+    // 1. importing a plain npm pacakge from a lazy engines does not add the package eagerly
+
+    assert.equal(requirejs.entries['fake-package'], undefined, 'fake-package should not be loaded before visting lazy engine');
+    await visit('/use-lazy-engine');
+    assert.equal(typeof requirejs.entries['fake-package'], 'object', 'fake-package was loaded only after visting the engine');
+    assert
+      .dom('[data-test-pkg-name]')
+      .hasText('fake-package', 'The fake-package was correctly imported');
+
+    // 2. Deps that are shared between app and engine share the same module state
```
ef4, line 219: We could write several more tests like this one that expose similar problems that remain un-addressed.

For example, if a dependency is used by two lazy engines but not the app, I think we will get the same duplicated module state problem.

ef4 (changes requested):
I didn't know about `dependsOn` and if we were dealing with only a well-known and hard-coded set of possible bundles it would probably be good enough.  But ember-engines is more flexible than that. Summarizing the key areas that I don't think work here:

1. Engines can use addons, which at some arbitrary deeper depth use ember-auto-import. Those dependencies need to get assigned to the right bundles. 

2. Engines can nest. They are not a flat space. So if we're using `dependsOn`, it needs to reflect that some engines depend on each other. Dependencies may hoist to multiple levels, not just the app vs the engine that directly used them.

3. Dependencies can get shared by arbitrary subsets of engines. They must not get duplicated module state even under those conditions.

Fixing this properly implies reimplementing a large piece of webpack. Which is why I don't think we should be doing it this way.

Instead, I still recommend a path where we use only *local* information about each addon to emit modules that webpack can analyze to perform *global* bundling and optimization.

thoov:
@ef4 Could you review this again?

thoov:
@ef4 friendly ping again on this

rwjblue (commented):
Chatted with @thoov in a quick call, and came up with a few changes:

* Rename `window.engineLookup` to `window._eai_engineLookup()`
* Remove the `treeForAddon` in favor of sending a PR to ember-asset-loader that tries our global if it exists (along with a comment linking to the code in ember-auto-import that sets it up)

thoov:
This will need to land first: https://github.com/ember-engines/ember-asset-loader/pull/127

ef4:
ember-asset-loader is released, but it is a transitive dependency of ember-engines. I tried to use NPM overrides here to get CI passing, but apparently NPM overrides doesn't actually work when you're also using workspaces. 😭https://github.com/npm/cli/issues/4834

ef4:
I am happy to land this but somebody needs to get on ember-engines to release.

lbdm44:
@ef4 I'm looking to get this pushed forward as we need this resolved for our application, I can create PRs as necessary to push this forward from my side.

I just want to confirm my understanding of what needs to happen:

1. The `ember-asset-loader` PR was merged and has landed in `1.0.0` of that repo. ✅ 
2. `ember-engines` needs to bump its dependency on `ember-asset-loader` to `1.0.0` since it is still on `0.6.1`
3. We can then check in this PR

runspired:
I can handle an ember-engines PR/release cc @SergeAstapov. I'll probably PR this over night tonight if you might have time to review tomorrow?

SergeAstapov:
> I can handle an ember-engines PR/release cc @SergeAstapov. I'll probably PR this over night tonight if you might have time to review tomorrow?

There is PR already, need also get https://github.com/ember-engines/ember-engines/pull/818 which I assume mo brained but would require v0.9.0

NullVoxPopuli:
looks like we're ready to move forward? 🎉

SergeAstapov:
@NullVoxPopuli yes, after we solve test failures in https://github.com/ember-engines/ember-engines/pull/820

lbdm44:
Looks like ember-engines/ember-engines#820 has gone in, any ideas on what needs to be done to get this ✅?

runspired:
the blocker in EmberEngines has been fixed and released as 0.9 https://github.com/ember-engines/ember-engines/releases/tag/v0.9.0

test-scenarios/package.json
```diff
@@ -36,6 +36,7 @@
     "ember-cli-typescript-2": "npm:ember-cli-typescript@^2.0.0-beta.3",
     "ember-cli-typescript-3": "npm:ember-cli-typescript@^3.0.0",
     "ember-cli-typescript-4": "npm:ember-cli-typescript@^4.0.0-alpha.1",
+    "ember-engines": "^0.8.20",
```
SergeAstapov, line 39: now should be `"ember-engines": "^0.9.0"` as v0.9.0 what includes support for ember-asset-loader@1.0.0

NullVoxPopuli:
should this be rebased? the diff seems sus:
image

esp as this PR didn't initially change any dependenices

runspired:
@NullVoxPopuli probably deps need recalculated but a diff is expected: https://github.com/ef4/ember-auto-import/pull/512/files#diff-fbd8a1b3659adac1c6d2d635ea65aa1cf0682939f6f767a03039c63ad5ca4f7bR39

thoov:
@rwjblue / @ef4 I rebased and get the tests suite to pass again. Any chance to land this and get a version released?

Kaceykaso:
@ef4 can we get eyes on this - I hear you are the only one that can release this change once merged?

lbdm44:
+1 This is blocking multiple projects at our company.

gabrielcsapo:
+1 would be great to have this merged and published.

ef4:
By `this` do you mean my last commit that turns off the `removeAvailableModules` optimization in development?

ef4:
I've been picking up this work, and the implementation here is still not correct. 

We shouldn't need to do any manual deduplication between test and app, because that's the point of the layering.

And we shouldn't be inserting any app chunks into the tests entrypoint, because again that's the point of layering them together inside code that webpack manages.

Hopefully I will have some time in the next couple weeks to try to fix this.

runspired:
@ef4 any progress here?

ef4:
Putting this on the embroider team agenda for tomorrow, with the intention of identifying who can pick this up and how soon. It's still a priority to get this fixed.

PR #516, opened 2022-04-11
Bump LTS versions + ember-cli-dependency-checker versions

thoov:
Working on CI failures with floating deps.

Bumping LTS version to 3.16


packages/addon-template/package.json
```diff
@@ -37,7 +37,7 @@
     "babel-eslint": "^10.1.0",
     "broccoli-asset-rev": "^3.0.0",
     "ember-cli": "~4.0.1",
-    "ember-cli-dependency-checker": "^3.2.0",
+    "ember-cli-dependency-checker": "3.2.0",
```
rwjblue, line 40:
```suggestion
    "ember-cli-dependency-checker": "^3.3.1",
```

rwjblue: The fix was landed to cli-dependency-checker

thoov:
@rwjblue / @ef4 could you review

ef4:
I think you're getting trolled by changes in either scenario-tester or Volta that are breaking our [per-scenario node version control](https://github.com/ef4/ember-auto-import/blob/44bd54db1ddf876ee542c73af651a1dadc395d4e/test-scenarios/scenarios.ts#L8-L17).

I think that is the source of the failures. We're supposed to be using old node to test old ember.

I would very much prefer to not do a semver major just to drop ember versions that are actually working fine.

ef4:
Attempting to fix in https://github.com/ef4/ember-auto-import/pull/517

ef4:
Superseded by #517.

PR #517, opened 2022-04-14
Fix our usage of volta in the test suite

ef4:
We had a workaround that allowed us to use volta to independently control the node version per scenario. That doesn't work on current volta anymore, but now there is a `volta run` command that gives us the same effect without a hack.


ef4:
I verified that this really does work on IE11, we're having unexplained CI infra problems with that test.

PR #518, opened 2022-04-18
Fix `style-loader` not being in effect in `production` environments

### LINKED ISSUE #514: Add ability to pass options to `MiniCssExtractPlugin`
In `ember-cli-build.js` in `autoImport` I have:

```js
styleLoaderOptions: {
  attributes: {
    nonce: 'style-loader',
  },
  insert: function (element) {
    document.getElementsByTagName('head')[0].prepend(element);
  },
},
```

Which works great on `ember-auto-import` 2.4.1 in development but in production this is not executed and hence all styles are inserted as links at the end of `head` (which is the default behavior of `style-loader` - i.e. the `insert` option is not used at all). The same happens in 2.4.0 (it also had another [issue](https://github.com/ef4/ember-auto-import/issues/493)). In 2.3.0 that is not the case and there all works fine.

cc @ef4

boris-petrov:
Fixes #514.

@ef4 - I believe the issue is [here](https://github.com/ef4/ember-auto-import/blob/v2.4.1/packages/ember-auto-import/ts/webpack.ts#L225) coming from [this](https://github.com/ef4/ember-auto-import/commit/788fe121630009d0b9c666761b8d7bd0ca586dc8) commit - in `production` environment `eai-style-loader` is not loaded at all. Not sure why this was done - this PR sets `style-loader` to load in all cases.

### DIFF
--- a/packages/ember-auto-import/ts/webpack.ts
+++ b/packages/ember-auto-import/ts/webpack.ts
@@ -86,6 +86,11 @@ window._eai_r = require;
 window._eai_d = define;
 `;
 
+type StyleLoader = {
+  loader: RuleSetUseItem;
+  plugin: WebpackPluginInstance | undefined;
+};
+
 export default class WebpackBundler extends Plugin implements Bundler {
   private state:
     | {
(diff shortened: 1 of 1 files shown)


boris-petrov:
I've made that `eai-style-loader` runs after `MiniCssExtractPlugin.loader` - not sure if that is correct. Is that why the tests are failing?

ef4:
It was intentional that we don't use style-loader in production. [style-loader's own docs](https://github.com/webpack-contrib/style-loader#recommend) say it's not recommend for production use and to use MiniCssExtractPlugin instead, and that you should never use both together simultaneously.

PR #520, opened 2022-05-18
Allow disabling sourcemaps when forbidEval is set (v1.x)

### LINKED ISSUE #444: Can't disable sourcemaps with forbidEval set
If `forbidEval` is set, it forces the webpack devtool value to be `'source-map'`. It does this after applying the user's custom webpack config, so the user doesn't have a way to specify something like

```js
forbidEval: true,
webpack: {
  devtool: false,
}
```

Is there a technical reason for this? If not, is ember-auto-import open to a PR changing the behavior here?

Also we are [people with unnecessarily-specific deployment code](https://github.com/ef4/ember-auto-import/blob/0fba64e4713ed93812822440ac8cd99679e2700b/docs/upgrade-guide-2.0.md#inject-entry-chunks-directly-in-indexhtml) and adopting ember-auto-import 2.0's dynamic chunk scripts will be a project for us, so we are stuck on 1.x for now. 😞 

Is 1.x still accepting patches?

apellerano-pw:
Fixes #444 for 1.x

I went with @ef4's advice to give precedence to the user's webpack config.

I tried to write a unit test for `Bundler.bundlerHook`, but the stubbing rabbit hole went too deep. Happy to try again if you have advice for how to test my change.

Aside:
When I tried to get local development set up for `tags/v1.12.1`, `npm i` would fail due to an integrity mismatch on ef4/node-fixturify-project. (See commit message for error log.) I healed the package-lock by deleting and re-adding the package, which did some minor bumps. I figure since you own both packages you can speak to the source & legitimacy of the integrity mismatch.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7747,6 +7747,7 @@
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/broccoli-source/-/broccoli-source-3.0.0.tgz",
       "integrity": "sha512-G4Zc8HngZIdASyQOiz/9H/0Gjc2F02EFwhWF4wiueaI+/FBrM9Ixj6Prno/1aiLIYcN0JvRC3oytN9uOVonTww==",
+      "dev": true,
       "dependencies": {
         "broccoli-node-api": "^1.6.0"
       },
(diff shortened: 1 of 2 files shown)


ef4:
Thanks, I cherry picked the changes onto #523, the integrity mismatch thing was not something I could reproduce.

ef4:
Published in 1.12.2.

ef4:
This is the corresponding fix for the 2.x series https://github.com/ef4/ember-auto-import/pull/524

PR #521, opened 2022-05-20
Adjust appRoot to account for dummy apps

ef4:
Dummy apps have a logical root that is different from their package root.

Mostly we are tracking the app root so that package resolution knows when devDependencies are relevant (they're relevant in the app and nowhere else). The classic build glosses over this distinction. But embroider cares and treats the dummy app as a distinct package, so if you get the app root wrong it can seed the shared PackageCache with the wrong value and cause the dummy app to not see its own devDependencies.

I think this is source of failures in https://github.com/simplabs/qunit-dom/pull/1423


ef4:
The failures on LTS are revealing a bug in fixturify-project, which I'm fixing in https://github.com/stefanpenner/node-fixturify-project/pull/70

PR #531, opened 2022-07-25
Improve upgrade-guide

pomm0:
Unfortunately, I missed this part as it looks very similar to `fingerprint.prepend`, but it really is important to not overlook the trailing `/assets` otherwise the url will be wrong. I wasted a couple hours with this and figured some others might miss that too. So probably a good reason to explicitly mention that.
Since I'm not a native speaker I would be happy if someone may give a better text suggestion


ef4:
Thanks!

PR #535, opened 2022-08-23
use node path.join to build os specific file name

void-mAlex:
avoids adding a duplicate analyzer-plugin to webpack on windows os

fixes an ERROR: Duplicate plugin/preset detected. 

Duplicated detected are:
"...node_modules\\ember-auto-import\\js\\analyzer-plugin.js",
"options": {
  "MARKER": "eaimeta@70e063a35619d71f"
}, ...


void-mAlex:
while there are 5 failed checks I believe they were filing already before any of my changes. please advise if there is anything else I can do on here to move this forward. thank you

ef4:
Thanks

PR #541, opened 2022-10-13
Respect v2 addon's explicit external list

ef4:
As a compatibility feature, v2 addons are allowed to list externals in the package metadata. These are dependencies that should be resolved at runtime from the surrounding environment, not at build time. Embroider respects this option but ember-auto-import does not.


ef4:
(As a general point, this kind of decision needs to move into `@embroider/shared-internals` so we're not constantly reproducing the same behavior in more than one place.)

Issue #519, opened 2022-05-16
Chunks changing content when updating Webpack?


njoyard:
Hi,

This is not really a bug report, but instead general questioning about how stuff works.

We have an app using `ember-auto-import` 2.4.1, and we recently had a production outage when deploying a webpack bump from 5.68.0 to 5.72.0. Insofar as we understand it as of this writing, the root cause was probably some cache layer not properly serving updated assets after deployment.

What happened is, after the webpack update, one of the ember-auto-import chunks changed contents but stayed with the same name. The only changes in the chunk are just reordering of object entries, eg:

```diff
1298c1298
< n.d(t,{_I:function(){return l},yW:function(){return c}})
---
> n.d(t,{yW:function(){return c},_I:function(){return l}})
```

The lack of filename change, combined with the very likely cache issue mentioned above and with our use of `ember-cli-sri`, meant that some users where served the updated `index.html` but the old chunk content (thus failing SRI).

We originally expected the filename would change, mistaking what is actually a chunkhash for a fingerprint. Tell me if I understood that part correctly, but it looks to me the chunkhash is just a hash of what the chunk contains in terms of modules and versions, while the fingerprint would be a hash of the actual final content of the file.

We're currently assessing steps to take in order to avoid future occurrences of this issue (one of them being, of course, understand _all_ our cache layers and fix them...), and we would like to maybe enable fingerprinting on those chunks. Is there some way of achieving that by configuring `ember-auto-import`? We're considering using a different chunk filename template, switching to `contenthash` or `fullhash` for example, but are those properly computed from the final content of the minified file?

Thanks for your insight on this :)

njoyard:
Thanks, I'll try that and report back.

davidtaylorhq:
> ```
> output: {
>  filename: 'chunk.[id].[contenthash].js'
> },
> ```
I tried this out in #548 to resolve the opposite problem - I was seeing identical chunk content, but the `chunkHash` was changing each time. Using `contenthash` did indeed make the `.js` filenames deterministic, but it revealed a separate problem.

The same filename template and contenthash is used for sourcemaps. So if you do something which changes the `.map` without changing the `.js` file, you'll end up with duplicate-named `.map` files with different content.

I think the module-naming improvements being discussed in #478 and #479 might help to improve the consistency of the `[chunkhash]`, which appears to change based on both the sourcemap and the output.

PR #548, opened 2022-11-07
Ensure chunk filenames are consistent for long-term caching

davidtaylorhq:
Per https://webpack.js.org/configuration/output/#template-strings, `contenthash` is based on the final content of the chunk file. Given the same file contents, the filename is guaranteed to be the same.

Resolves https://github.com/ef4/ember-auto-import/issues/519

### DIFF
--- a/packages/ember-auto-import/ts/webpack.ts
+++ b/packages/ember-auto-import/ts/webpack.ts
@@ -154,8 +154,8 @@ export default class WebpackBundler extends Plugin implements Bundler {
       output: {
         path: join(this.outputPath, 'assets'),
         publicPath: this.opts.publicAssetURL,
-        filename: `chunk.[id].[chunkhash].js`,
-        chunkFilename: `chunk.[id].[chunkhash].js`,
+        filename: `chunk.[id].[contenthash].js`,
+        chunkFilename: `chunk.[id].[contenthash].js`,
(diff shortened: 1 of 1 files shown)


ef4:
Thanks. Since I see you've already tried this in discourse, can you check if there's any impact on the development-time rebuild performance?

Just asking because if it turns out to be more expensive, we can use it for prod but not dev. If the cost doesn't change, this PR will be fine as is and we can just always use it.

davidtaylorhq:
Thanks for taking a look @ef4 

We didn't notice any performance change, but I haven't done any statistically-significant measurements. I did note that Webpack only [enables `realContentHash`](https://github.com/webpack/webpack/blob/9fcaa24357/lib/config/defaults.js#L1158) by default in production mode, which should help dev performance.

That said, after some more experimenting, I don't think this change is sufficient by itself. I've expanded a little in https://github.com/ef4/ember-auto-import/issues/519#issuecomment-1306334170. We've ended up disabling auto-import sourcemaps in production for now to improve cross-build caching.

I'm hopeful that #479 will help with the problem we're seeing. I think removing the ephemeral paths like `.../broccoli-58592ChasS63D8jsg/cache-294-webpack_bundler_ember_auto_import_webpack/...` from the sourcemaps should help with hash consistency across builds.

Issue #504, opened 2022-02-16
v2 addons don't always get ideal execution order


simonihmig:
I pulled my hair with some very broken behavior in an app. Luckily I was able to create a [minimal reproduction](https://github.com/simonihmig/broken-import-reproduction). So this is a super weird issue, be prepared...

## Context

I believe this has to do with the following circumstances:
* a v2 addon (in this case [tracked-redux](https://github.com/simonihmig/tracked-redux)) imports a v1 addon ([ember-tracked-storage-polyfill](https://github.com/ember-polyfills/ember-tracked-storage-polyfill))
* the app imports the v2 addon (`tracked-redux`, directly or through some other addon), and it also(!) imports the v1 addon (`ember-tracked-storage-polyfill`, again directly or through some other addon)
* during module resolution, the import of the v2 addons happens *before* the other v1 addon is imported through the app or some other addon. So the v1 addon (`ember-tracked-storage-polyfill`) module is loaded (by loader.js) the first time through the v2 addon.

> Note that this reproduction uses `tracked-redux`, as this is also the v2 addon used in the real app, and I didn't want to have to create another (simpler) v2 addon just for the reproduction. But I strongly believe the error here is not related to redux or something else in that v2 addon, as it doesn't do anything unusual, with regards to imports or anything related. AFAICT.

## Reproduction

* In the application route, `tracked-redux` is [imported](https://github.com/simonihmig/broken-import-reproduction/blob/master/app/routes/application.js#L3)
* `tracked-redux` imports `getValue` from `ember-tracked-storage-polyfill` and assigns it to an alias `consumeTag` [here](https://github.com/simonihmig/tracked-redux/blob/master/packages/tracked-redux/src/-private/tracking.ts#L15)
* However at the time the alias is assigned, the imported module of `ember-tracked-storage-polyfill` (webpack-require) is an empty object, so `consumeTag` is undefined. See this screenshot while debugging:
![Bildschirmfoto 2022-02-16 um 11 36 54](https://user-images.githubusercontent.com/1325249/154251932-d1526b64-d1fa-41ff-a8ec-28379e57c66d.png)
* when debugging into that webpack-require call, and following the wrapper module of eai that calls again into Ember's AMD-like `require` by loader.js, you can see that the module is in pending state, and indeed the returned `mod.module.exports` is just an empty object:
![image](https://user-images.githubusercontent.com/1325249/154252397-9f13815f-13f6-4a76-8d44-d8ed084875c6.png)
[...]

PR #553, opened 2022-12-08
Add v1 addons that could be consumed by v2 addons in module-scope to …

NullVoxPopuli:
…a list that loads them earlier in AMD

Partially resolves: https://github.com/ef4/ember-auto-import/issues/504


ef4:
Next step here is to remove things from the early boot set when their corresponding real package isn't in the app's package.json dependencies / devDependencies. This comes up in the test suite because our oldest supported ember doesn't support `@glimmer/tracking`.

NullVoxPopuli:
The fun deepens.
on that ember-source 3.4 + fastboot scenario, we get (after removing `@glimmer/tracking` from the early boot set because it is not present in the package.json):
```
loader.js:247 Uncaught Error: Could not find module `@ember/service/index` imported from `__v1-addons__early-boot-set__`
```

NullVoxPopuli:
I disabled the feature entirely for old ember.

NullVoxPopuli:
Green, sans ie11 timeout

PR #562, opened 2023-01-26
Prevent HTML-escaping of module specifiers

simonihmig:
Importing a file like `image.jpg?widths=50,100,640&formats=original,webp,avif&responsive` for consumption with a custom webpack loader failed, as what got written into the AMD shim was instead `image.jpg?widths&#x3D;50,100,640&amp;formats&#x3D;original,webp,avif&amp;responsive`, failing the webpack module rules and/or the parsing of the query in the loader.

It turned out this was due to the double curlies in the handlebars-based AMD shim generator HTML-escaping the module specifiers. With "normal" paths this would not have any effect, but here it breaks things obviously.

---

I haven't run tests locally, as I was expected this to not break anything, and I hoped Github would do that for me, but it seems CI setup is not working correctly (matrix scenarios not running, the `discover_matrix` [fails here](https://github.com/ef4/ember-auto-import/pull/562/checks#step:7:18)).

Also I didn't add any tests, as I wasn't sure how to do this, without replicating a whole e2e scenario with a webpack plugin reproducing the failure...


ef4:
CI looks good now, the IE11 test is flaky and I'm not concerned about it here.

simonihmig:
@ef4 what do you think about this? Can we merge? Or do we need a test?

ef4:
Looks good, thanks.

Issue #564, opened 2023-01-30
Expected hash or Mixin instance, got [object Undefined] after upgrading to 2.6.0


oliverlj:
I don't know where is the problem
When I start I have this in the console with a white page

```
index.js:106 Uncaught Error: Assertion Failed: Expected hash or Mixin instance, got [object Undefined]
    at assert (index.js:106:13)
    at buildMixinsArray (mixin.js:509:13)
    at Mixin.reopen (mixin.js:435:38)
    at DebugFrameworkObject.extend (core.js:370:12)
    at Module.callback (core_view.js:13:40)
    at Module.exports (loader.js:106:1)
    at Module._reify (loader.js:143:1)
    at Module.reify (loader.js:130:1)
    at Module.exports (loader.js:104:1)
    at Module._reify (loader.js:143:1)
```

When I start the test, it says :
```
not ok 3 Chrome 109.0 - [undefined ms] - Global error: Uncaught Error: Could not find module `@ember/test-helpers` imported from `cryptofiscafacile-gui/tests/test-helper` at http://localhost:7357/assets/vendor.js, line 266
    ---
        browser log: |
            {"type":"error","text":"Uncaught Error: Could not find module `@ember/test-helpers` imported from `cryptofiscafacile-gui/tests/test-helper` at http://localhost:7357/assets/vendor.js, line 266\n","testContext":{"state":"complete"}}
           ```

jelhan:
Several users of ember-bootstrap report seeing this error after upgrading to v2.6.0. I was able to trace it down to this line of code deep in Ember's internals:

```js
define("@ember/-internals/views/lib/views/core_view", ["exports", "@ember/-internals/metal", "@ember/-internals/runtime", "@ember/object/evented", "@ember/object/-internals", "@ember/-internals/views/lib/views/states"], function (_exports, _metal, _runtime, _evented, _internals, _states) {
  "use strict";

  Object.defineProperty(_exports, "__esModule", {
    value: true
  });
  _exports.default = void 0;
  var __decorate = void 0 && (void 0).__decorate || function (decorators, target, key, desc) {
    var c = arguments.length,
      r = c < 3 ? target : desc === null ? desc = Object.getOwnPropertyDescriptor(target, key) : desc,
      d;
    if (typeof Reflect === "object" && typeof Reflect.decorate === "function") r = Reflect.decorate(decorators, target, key, desc);else for (var i = decorators.length - 1; i >= 0; i--) {
[...]

PR #566, opened 2023-02-13
omit dependencies bundled with ember-source from participating in the webpack fun 

NullVoxPopuli:
Resolves: https://github.com/ef4/ember-auto-import/issues/564
Eliminates, what I believe to be a require-cycle:
- rsvp is require'd
- rsvp previously depended on the early boot set
- but the early boot set contains dependencies which depend on rsvp (`@ember/object` / mixins, etc)

-------

further investigations after the embroider office hours:

 - is rsvp duplicated?
   - `yarn why rsvp`
      ```
      ❯ yarn why rsvp
      yarn why v1.22.19
      [1/4] 🤔  Why do we have the module "rsvp"...?
      [2/4] 🚚  Initialising dependency graph...
      [3/4] 🔍  Finding dependency...
      [4/4] 🚡  Calculating file sizes...
      => Found "rsvp@4.8.5"
      info Has been hoisted to "rsvp"
      info Reasons this module exists
         - Hoisted from "ember-bootstrap#rsvp"
         - Hoisted from "ember-cli-typescript#rsvp"
      ```    
      yes
- why is rsvp used in `ember-bootstrap`?
   - no types, so the types from rsvp are not used
   - used in node for tests, so dep must exist so node knows where to find it
   - in browser (the addon code), the rsvp-shipped-with-ember would be sufficient      
- does `autoImport.exclude` with `rsvp` fix the problem?
  - from the app: no
  - from ember-bootstrap: yes
    ```js
    // node_modules/ember-bootstrap/index.js
    options: {
      '@embroider/macros': {
        setOwnConfig: {},
      },
      autoImport: {
        exclude: ['rsvp'],
      }
    }
     ``` 

**Path forward**
  ensure that [these dependencies](https://github.com/emberjs/ember.js/pull/20349/files#diff-92bbac9a308cd5fcf9db165841f2d90ce981baddcb2b1e26cfff170929af3bd1R169) are always excluded from ember-auto-import (at least until ember-source becomes a real package)

### DIFF
--- a/packages/ember-auto-import/ts/auto-import.ts
+++ b/packages/ember-auto-import/ts/auto-import.ts
@@ -1,3 +1,4 @@
+import fse from 'fs-extra';
 import Splitter from './splitter';
 import { Bundler, debugBundler } from './bundler';
 import Analyzer from './analyzer';
@@ -25,6 +26,7 @@ import semver from 'semver';
 import type { TransformOptions } from '@babel/core';
 import { MARKER } from './analyzer-syntax';
 import path from 'path';
+import walkSync from 'walk-sync';
 
 const debugTree = buildDebugCallback('ember-auto-import');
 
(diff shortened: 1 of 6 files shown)


test-scenarios/package.json
```diff
@@ -27,6 +27,8 @@
     "leader-v1": "npm:ember-auto-import@1.7",
     "leader-v2": "npm:ember-auto-import@1.11",
     "old-analyzer": "npm:ember-auto-import@2.1.0",
+    "ember-bootstrap": "^5.0.0",
```
NullVoxPopuli, line 30: I'd like to not have these added to test-scenarios, because it hides the minimal problem behind another dep, but I've yet to create a reproduction without -- stay tuned.

packages/ember-auto-import/ts/auto-import.ts
```diff
@@ -115,11 +116,40 @@ export default class AutoImport implements AutoImportSharedAPI {
...
+     */
+    let alreadyBundled = new Set();
+
+    emberSource: if (this.rootPackage.hasDependency('ember-source')) {
+      let emberPackageJsonPath = resolvePackagePath(
+        'ember-source',
+        this.rootPackage.root
+      );
+
+      if (!emberPackageJsonPath) break emberSource;
```
ef4, line 136: Let's not use labeled break when if statements would do just as well.

NullVoxPopuli: Updated, thanks!

NullVoxPopuli:
All green except for an IE11 test

jelhan:
Thanks a lot for working on this. Please let me know if there is anything I could do to get it released.

simonihmig:
As the issue has been resolved otherwise, this seems to be not needed anymore? @NullVoxPopuli can we close it?

NullVoxPopuli:
yes! thanks for reminding me!

PR #568, opened 2023-02-22
Change default boot set

NullVoxPopuli:
In the wake of https://github.com/ef4/ember-auto-import/issues/564 and the complexity introduced by: https://github.com/ef4/ember-auto-import/pull/566:

This PR inverts the default (intended bug fix, due to the bugs introduced by https://github.com/ef4/ember-auto-import/pull/553)

For folks that need the early boot set, you'll need to opt in.

The `earlyBootSet` is still a function, and still receives the default recommended values:
```js
autoImport: {
  earlyBootSet(defaults) {
    return [...defaults];
  }
}
```

cc @simonihmig @vitch


NullVoxPopuli:
@ef4 the tests that work with the earlyBootSet set the option manually. All existing tests are passing -- lemme know if you think more is needed

jelhan:
Thanks a lot.

PR #573, opened 2023-03-23
added babel-plugin-ember-template-compilation

candunaj:
ember-auto-import is used for loading V2 addons. There is a bug, that babel-plugin-htmlbars-inline-precompile is always used. For newer versions of ember-source>=3.27.0 babel-plugin-ember-template-compilation should be used.

This PR can be the start of a discussion of how to solve it. If there is a suggestion on how to improve the solution, I would like to implement it.


ef4:
Thanks this looks good. Please see lint failures in CI.

ef4:
Thanks. 

The ci failures for canary are unrelated and already fixed on main.

PR #574, opened 2023-04-03
Detect where modules loop back to AMD from webpack

### LINKED ISSUE #504: v2 addons don't always get ideal execution order

ef4:
Fixes #504.

Removes earlyBootSet, because that was a hack for the lack of this feature. earlyBootSet never actually worked correctly anyway. I'm considering this a bugfix and not a breaking change, since nobody could reliably use that setting anyway.


ef4:
Next steps here:
 - ember canary tests are failing, I think not because of this PR
 - fastboot tests are failing because we need to apply the deps rewriting to some fastboot-specific code that is being missed

ef4:
There is one remaining test failure, which has to do with user-configured custom webpack externals. We need to handle those differently than the externals emitted from ember-auto-import.

mansona (approved):
amazing 😍

PR #579, opened 2023-05-03
Tweak watchedDirectories behavior

NullVoxPopuli:
The problem this PR aims to solve is that when working with `watchedDependencies`, _any change_ to the entire directory of the watched dependency causes the app to rebuild. Since an app only really cares about the shipped version of the dependency, if that dependency specifies a `files` entry in its package.json, folders listed there will be used instead of the package-root.

For example:

```
autoImport: {
  watchedDependencies: ['my-library'],
}
```

_today_:
```
node_modules/my-library/
  all of this is watched
  src
  dist
  package.json
```

_with this PR_
```
node_modules/my-library/
  dist
    only this sub directory is watched
```

There is a downside though -- during development, the package.json contents can still change, and would be relevant to the consuming app.
I think this is an acceptable tradeoff tho, because ember-cli is bad at handling dependency changes in node_modules anyway, and would need re-starting. Where we lose needed behavior is when other static entries in package.json are read by tools like auto-import, like `ember-addon.app-js` -- this object is updated as you add/remove app-re-exports from v2 addons.

As implemented, this would be a breaking change, so maybe it'd be better to have a separate autoImport option for this sort of behavior? I think so -- but what would it be called? `autoImport.watchPackages`?

### DIFF
--- a/packages/ember-auto-import/ts/package.ts
+++ b/packages/ember-auto-import/ts/package.ts
@@ -1,6 +1,6 @@
 import resolvePackagePath from 'resolve-package-path';
 import { join, dirname } from 'path';
-import { readFileSync } from 'fs';
+import { lstatSync, readFileSync } from 'fs';
 import { Memoize } from 'typescript-memoize';
 import type { Configuration } from 'webpack';
 import {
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
I don't feel like working on this, and the answer is probably to fix dependenciesMeta*injected, so that they can be default

PR #581, opened 2023-05-17
Add link from top level changelog to actual changelog

fhightower:
Hi there! I suggest updating the top level changelog to link to the actual changelog. Minor, but it makes life easier for readers looking for the changelog. Thanks!


CHANGELOG.md
```diff
@@ -1 +1 @@
-packages/ember-auto-import/CHANGELOG.md
\ No newline at end of file
+[packages/ember-auto-import/CHANGELOG.md](https://github.com/embroider-build/ember-auto-import/blob/main/packages/ember-auto-import/CHANGELOG.md)
```
void-mAlex, line 1: this should be a relative link
current pr ties it to github.com so it's not usable on a checked out repo to jump to file, as well as always pointing to the main branch which I'm not convinced is always the right thing to be doing

PR #584, opened 2023-06-12
Enable `bugfixes: true` to Babel's config

boris-petrov:
That leads to (much) smaller code and will be the default in Babel 8. I got to that because my `targets` contains `safari >= 16` but due to [this](https://github.com/babel/babel/issues/13916) code was transpiled for much older browsers. I fixed the problem for classic packages in Ember (by adding `bugfixes: true` to `babel` in `ember-cli-build.js`) but for `ember-auto-import`ed ones I couldn't - this flag is needed for them.

@ef4 have I done the correct thing here?

### DIFF
--- a/packages/ember-auto-import/ts/package.ts
+++ b/packages/ember-auto-import/ts/package.ts
@@ -575,6 +575,7 @@ export default class Package {
         [
           require.resolve('@babel/preset-env'),
           {
+            bugfixes: true,
             modules: false,
             targets: parent.targets,
           },


ef4:
Thanks, this seems like a good idea, and from reading the description of the `bug fixes: true` option is seems pretty safe to enable as a non-breaking change.

I think the test failures here are happening because some of the stuff being left un-transpiled now doesn't work in the node version we're testing under. It may be that we just need to add node to the targets for those test cases.

boris-petrov:
@ef4 thanks for the review! Do you want me to do it or are you? I'll need a pointer or two if it's going to be me - what should be changed and where. :smile:

PR #586, opened 2023-07-13
Feature: allowAppImports

ef4:
This is work toward a new feature that would allow apps to opt-in to letting ember-auto-import manage parts of the app itself.


mansona:
I'm going to take over this PR so I'll close it and re-create it (so @ef4 can review 😂 )

PR #587, opened 2023-07-13
Feature: allowAppImports

mansona:
This is taking over from https://github.com/embroider-build/ember-auto-import/pull/586 👍


ef4:
TODO list items:
 - [ ] filtering matching files out of classic build pipeline. May involve monkey-patching the appInstance's trees.app.
 - [ ] teaching the splitter to emit imports for package-name-prefixed imports that match, like "my-app/lib/example1".
 - [ ] teaching the splitter to emit imports for relative imports for files that match, like "../lib/example1"
 - [ ] handling webpack externals so that when an auto-imported module imports a classically-built app module it doesn't get duplicated inside webpack. This is an extension to what we already do to support v2 addons using ember-provided features.
 - [ ] make babel-loader config sensitive to the fact that app code gets the full app babel config (whereas dependencies get a streamlined, standardized config)

packages/ember-auto-import/ts/package.ts
```diff
@@ -481,6 +482,15 @@ export default class Package {
     }
   }
 
+  get allowAppImports(): string[] {
+    // only apps (not addons) are allowed to set this
```
simonihmig, line 486: I discussed this feature a while ago with @ef4 within CrowdStrike, and IIRC we came to the conclusion that also addons should be able to do this. Example use case would be to have a v1 addon import its own css-modules based CSS (to get rid of ember-css-modules), so the consuming app could already switch to optimized Embroider (which ember-css-modules does not support) without requiring all addons to migrate to v2.

FWIW, this (v1 addon support of this feature) is not anymore a priority for _us_, but maybe it's still good to allow this!?

ef4: This is a good point. 

I want to use this feature to make a package like https://github.com/cardstack/glimmer-scoped-css work in classic builds, and for that story to be complete you'd want a v1 addon to be able to use it too.

packages/ember-auto-import/ts/auto-import.ts
```diff
@@ -75,6 +79,20 @@ export default class AutoImport implements AutoImportSharedAPI {
     this.consoleWrite = (...args) => addonInstance.project.ui.write(...args);
   }
 
+  installAppFilter(_host: AppInstance) {
+    // TODO upstream this type change to @embroider/shared-internals
+    let host: AppInstance & {
+      trees: {
+        app: Node;
+      };
+    } = _host as any;
```
MrChocolatine, line 88: Since you force the type, could this work/be considered:

```suggestion
    let host = _host as AppInstance & {
      trees: {
        app: Node;
      };
    };
```

PR #596, opened 2023-10-24
Support private properties and static blocks

andreyfel:
We need to add `@babel/plugin-transform-class-static-block` and `@babel/plugin-proposal-private-methods` to the list of our plugins to correctly process the code with private properties and static blocks.

### DIFF
--- a/packages/ember-auto-import/ts/package.ts
+++ b/packages/ember-auto-import/ts/package.ts
@@ -537,8 +537,13 @@ export default class Package {
 
     let plugins = [
       [require.resolve('@babel/plugin-proposal-decorators'), { legacy: true }],
+      [require.resolve('@babel/plugin-transform-class-static-block')],
       [
         require.resolve('@babel/plugin-proposal-class-properties'),
+        { loose: false },
+      ],
+      [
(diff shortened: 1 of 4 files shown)


test-scenarios/doubly-indirect-test.ts
```diff
@@ -51,8 +51,25 @@ function makeAddon() {
   addon.addDependency('some-lib', {
     files: {
       'index.js': `
+
+    class ClassWithStaticBlockAndPrivateProperties {
```
andreyfel, line 55: Not sure if it is the best place to test that. Please let me know if there is a better place. Maybe a new test file?

packages/ember-auto-import/ts/package.ts
```diff
@@ -537,10 +537,15 @@ export default class Package {
 
     let plugins = [
       [require.resolve('@babel/plugin-proposal-decorators'), { legacy: true }],
+      [require.resolve('@babel/plugin-transform-class-static-block')],
```
andreyfel, line 540: For some reason, the test passes even without this plugin...

mansona (approved):
Looks good 🎉 we discussed this in the triage meeting today and we're happy to merge 👍

PR #598, opened 2023-11-24
Add release-plan for automating releases

mansona:
This uses [release-plan](https://github.com/embroider-build/release-plan) which was originally developed for [embroider](https://github.com/embroider-build/embroider) and recently extracted into its own package.

This PR adds the workflows that create the Release PR and the publish step 👍

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,53 @@
+# For every push to the master branch, this checks if the release-plan was
+# updated and if it was it will publish stable npm packages based on the
+# release plan
+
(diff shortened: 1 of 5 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,53 @@
+# For every push to the master branch, this checks if the release-plan was
```
NullVoxPopuli, line 1:
```suggestion
# For every push to the main branch, this checks if the release-plan was
```

NullVoxPopuli (approved):

PR #602, opened 2023-11-28
Allow arbitrary extensions for app-imports

simonihmig:
With a config like `allowAppImports: ['images/**/*.webp']` and importing a file like `image.webp`, eai would fail to handle that import properly, as it assumes that explicit extensions would be 1-3 characters long.

With a proper webpack loader we can make it import any type of file, so we shouldn't make assumptions about the possible extensions.


mansona (approved):
seems reasonable to me 👍

PR #603, opened 2023-12-03
Fix imports with a query part

simonihmig:
With the new `allowAppImports` feature, apps can import resources that are handled with custom webpack loaders. Some can allow you to customize their behavior by passing query params, like `import image from './images/image.jpg?size=1024`.

This PR is fixing two related issues:
* make sure that an `allowAppImports` glob like `'images/**/*.jpg'` would also match for imports that include query params, by stripping the query part before trying to match the import to the glob
* query params could technically also include quotes. These already got escaped when building the app-chunk that has the AMD-shims, but the regex to replace `EAI_DISCOVERED_EXTERNALS` was not taking their possible existence into account correctly, leaving `EAI_DISCOVERED_EXTERNALS` around in the final build.

### DIFF
--- a/packages/ember-auto-import/ts/package.ts
+++ b/packages/ember-auto-import/ts/package.ts
@@ -13,6 +13,7 @@ import semver from 'semver';
 import type { TransformOptions } from '@babel/core';
 import { MacrosConfig } from '@embroider/macros/src/node';
 import minimatch from 'minimatch';
+import { stripQuery } from './util';
 
 // from child addon instance to their parent package
 const parentCache: WeakMap<AddonInstance, Package> = new WeakMap();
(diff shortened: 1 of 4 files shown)


simonihmig:
I haven't yet looked into adding tests for this. I guess for covering imports of non-JS files, including query params, we would need to add a simple webpack loader in a scenario test? Like on that returns the query as the generated module's export, so we can assert that the loader is correctly invoked and the query stuff works? Should that be a separate scenario? Or any other suggestions?

simonihmig:
Btw, I worked on these changes as part of making https://github.com/simonihmig/ember-responsive-image/pull/442 work with ember-auto-import, and it seems this is indeed the final missing piece, as the things that worked under Embroider are now also passing (locally) under eai! :tada:

simonihmig:
Added tests, this should be good now for another review! @ef4 @mansona

ef4 (approved):

PR #613, opened 2024-03-08
fix imports analysis of gts

### LINKED ISSUE #611: issue with gts/gjs if ts transform is enabled
Since ts babel plugin removes the unused (but used in template) imports at program enter, this plugin doesn't have a chance to analyse them.

This plugin probably needs to do the same as ember babel template compilation.
Or just analyse the imports at program exit 

Maybe the approach needs some rethink as every plugin that wants to analyse imports of gts files needs to adop the same behaviour.

patricklx:
We want to analyse imports that are dynamically added by babel plugins. We cannot use Program:exit, because
amd module transform removes the imports at that time.
Therefore, we add an empty statement with extra attributes at the end of the program body.
When we reach it, we run the Import analysis

fixes #611


patricklx:
this is too hacky, i think there are better solutions in babel plugin ember compilation

PR #629, opened 2024-06-22
only check devDependencies when checking requested range of an app package

mansona:
I noticed this when I was linking a package that had a peerDependency range wider than it's devDependencies (ember-showdown-prism)

i don't understand why we would ever want to check the requested range of a dev depdency of a package 🤔 but perhaps there is a nuance that I don't understand here and we need to encode that nuance into the question that is being asked

### DIFF
--- a/packages/ember-auto-import/ts/package.ts
+++ b/packages/ember-auto-import/ts/package.ts
@@ -254,7 +254,6 @@ export default class Package {
     let { pkg } = this;
     return (
       pkg.dependencies?.[packageName] ||
-      pkg.devDependencies?.[packageName] ||
       pkg.peerDependencies?.[packageName]
     );
   }
--- a/packages/ember-auto-import/ts/package.ts
+++ b/packages/ember-auto-import/ts/package.ts
@@ -252,9 +252,15 @@ export default class Package {
   // package.json
   requestedRange(packageName: string): string | undefined {
     let { pkg } = this;
-    return (
-      pkg.dependencies?.[packageName] || pkg.peerDependencies?.[packageName]
-    );
+    let result =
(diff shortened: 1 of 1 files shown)


mansona:
Turns out fixing a lint error is very tough on the mobile app 🤣 I am very proud of my achievement 💪

ef4:
I think this code applies to apps too though. Which need their devDependencies.

We also have `hasNonDevDepDependency()` in this same class. Code like `assertAllowedDependency` is sensitive to the difference between apps and addons. If you found a spot that isn't, that is where the bug should be fixed.

mansona:
I started trying to write down some context for this change and then I figured it would be better to just fix it rather than writing it down 😂  the `requestedRange` function is only used in one place so it was quite simple to add the new arg to it 👍 

I'm pretty sure this fixes my issue but I'm wondering if you could help me place a test for this (unless you think it's not necessary?) 🤔

packages/ember-auto-import/ts/package.ts
```diff
@@ -250,13 +250,19 @@ export default class Package {
 
   // the semver range of the given package that our package requests in
   // package.json
-  requestedRange(packageName: string): string | undefined {
+  requestedRange(
+    packageName: string,
+    includeDevDeps: boolean
```
ef4, line 255: This shouldn't be an argument. `includeDevDeps === !this.isAddon`

mansona: 🙈 I really should have noticed I was inside the package here 😫

mansona: fixed 👍

ef4 (approved):

PR #630, opened 2024-07-04
Update TypeScript

simonihmig:


packages/ember-auto-import/ts/webpack.ts
```diff
@@ -32,8 +32,8 @@ const EXTENSIONS = ['.js', '.ts', '.json'];
 
 const debug = makeDebug('ember-auto-import:webpack');
 
-registerHelper('js-string-escape', jsStringEscape);
-registerHelper('join', function (list, connector) {
+Handlebars.registerHelper('js-string-escape', jsStringEscape);
+Handlebars.registerHelper('join', function (list, connector) {
```
simonihmig, line 36: The API docs only show that kind of usage, and `registerHelper` accesses `this.`. Not sure why it worked before, but as soon as the compilation output from a new TS versions changed this be like `(0, handlebars_1.registerHelper)('join', ...)`, it would lose access to `this`.

packages/ember-auto-import/ts/analyzer.ts
```diff
@@ -192,7 +192,7 @@ export default class Analyzer extends Funnel {
       try {
         transformSync(source, options);
       } catch (err) {
-        if (err.name !== 'SyntaxError') {
+        if (err instanceof Error && err.name !== 'SyntaxError') {
```
mansona, line 195: I understand what you're doing here but I don't like the idea of introducing a subtle behaviour change just to work around types here. I would prefer if you worked around the  types like you did later in the PR. Something like `(err as unknown as {name: string})`? but don't quote me on it 😂

simonihmig: But what's the change in behaviour here? In case of a `SyntaxError` it will always be an instance of `Error`, right? And if it's not a `SyntaxError`, then the if condition will not be hit either way. If feel this is the cleaner way to do this than to override the types, given that TS is actually right here.

simonihmig: @mansona 👆😏

ef4 (approved):

PR #659, opened 2025-06-04
Disable ember-auto-import in vite projects

NullVoxPopuli:
This prevents the error:
image

when using vite without ember-auto-import/webpack declared in your package.json.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.DS_Store
 packages/*/node_modules
 packages/*/tmp
 packages/*/dist
--- a/packages/ember-auto-import/ts/index.ts
+++ b/packages/ember-auto-import/ts/index.ts
@@ -8,10 +8,49 @@ module.exports = {
 
   init(...args: any[]) {
     this._super.init.apply(this, args);
+
+    if (!this.isEnabled()) {
+      return;
+    }
+
     AutoImport.register(this);
   },
 
+  __host: null,
+  eldestGrandparent() {
+    if (this.__host !== null) return this.__host;
+
+    let parent = this.parent;
+    while (parent) {
+      if (!parent.parent) break;
+      parent = parent.parent;
+    }
+    this.__host = parent;
+    return this.__host;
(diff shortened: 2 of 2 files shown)


NullVoxPopuli:
1. To remove ember-auto-import from compat vite apps
2. This was suggestion from @mansona, previously, i was detecting if vite and @embroider/vite were present in the host app but @mansona said that also wasn't good to do. What _should_ be the way to disable auto-import found in v1 addons present in vite-using apps? (Entirely in embroider? As you say in 3)
3. Do you mean auto-import v1? So far, i've found it safe to override to v2, which with most addons declaring ^2.something this pr would work.... But! I didn't know embroider could just.... Patch auto-import's infra itself. That _does_ seem better, and the embroider repo is easier to add test projects to (due to the existing templates over there)

I'm gonna close this and pr over there

NullVoxPopuli:
Re-opening, because in an attempt to solve this in embroider, here: https://github.com/embroider-build/embroider/pull/2507/files

we (@mansona and I) found that none of embroider's infra is involved when this error occurs
```
To use these addons, your app needs ember-auto-import >= 2: ember-bootstrap, ember-popper-modifier, ember-source
```

This can be traced by 
```js
'use strict';

const EmberApp = require('ember-cli/lib/broccoli/ember-app');
const { compatBuild } = require('@embroider/compat');

module.exports = async function (defaults) {
  const { buildOnce } = await import('@embroider/vite');
  // breakpoint here
  let app = new EmberApp(defaults, {});

  return compatBuild(app, buildOnce);
};
```
then we find ourselves here where `included` is called:
```
this.project.addons.forEach((addon) => {
  if (addon.included) {
    // error happens here
    addon.included(this); // where auto-import's leader is chosen
  }
});
```

The only way embroider could get involved with auto-imports leader-chooser is to still make a change to auto-import _anyway_ to be aware of embroider as a leader.
But! We don't want ember-auto-import to do anything. It needs to no-op.

PR #661, opened 2025-06-23
introduce `entrypointInjectables` to allow configuring webpack vars

fivetanley:
Allows configuring magic webpack vars like `__webpack_nonce__` and `__webpack_public_path__`. These need to be set in each entry point file created by Ember CLI. Unfortunately, this can't be set in `app.js` or via `app.import`. Configuration of these vars is necessary for configuring Content Security Policy under certain scenarios or setting the hostname for a CDN that can only be known at runtime instead of build time.

Usage would look something like this in the ember-cli-build.js file:

```javascript
autoImport: {
  entrypointInjectables: ['__webpack_public_path__ = window.__PUBLIC_PATH__']
}
```

Then, in the index.html manipulated server side that sets that var _before_ the ember-auto-import inject'd tags are loaded:

```html
<script>
  window.__PUBLIC_PATH__ = 'https://mycdnaddress.com/';
</script>
```

addresses https://github.com/embroider-build/ember-auto-import/issues/409


ef4:
Putting `__webpack_public_path__ = ` into `app.js` doesn't work because webpack isn't processing app.js. But we already have an option to designate which files in the app you want webpack to process. For example, if you set:

```js
  autoImport: {
      allowAppImports: ['lib/**'],
  },
```

Then webpack is in charge of all of `app/lib`, and you can make a file in there like:

```js
// app/lib/public-path.js
__webpack_public_path__ = 'http://example.com/';
```

And import that file first from `app.js`:

```js
// app.js
import './lib/public-path'
// ...
```

And from my quick testing this is enough to point webpack at a new public path.

Please give that a try, because I think it means we don't need this new option.

PR #664, opened 2025-09-15
add caching layer for discovered external chunks

### LINKED ISSUE #577: Modifying SASS files triggers full reload since 2.6.2
Hi there, 

modifying a SASS file causes full page reload instead of CSS livereload in Ember since ember-auto-import 2.6.2. It wasn't the case when I was using 2.6.1 previously.

Thx

apellerano-pw:
Trying the idea from https://github.com/embroider-build/ember-auto-import/issues/577#issuecomment-1568588073 to Fix #577 

Seems to work locally. Is caching the input OK or should it be the output?

### DIFF
--- a/packages/ember-auto-import/ts/webpack.ts
+++ b/packages/ember-auto-import/ts/webpack.ts
@@ -135,6 +135,8 @@ export default class WebpackBundler extends Plugin implements Bundler {
 
   private lastBuildResult: BuildResult | undefined;
 
+  private outputCache = new Map<string, string>();
+
   constructor(priorTrees: InputNode[], private opts: BundlerOptions) {
     super(priorTrees, {
       persistentOutput: true,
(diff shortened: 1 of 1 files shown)


apellerano-pw:
I've noticed that the very first time I save an SCSS file after running `ember s`, I'll still get a reload triggered. I guess the first time build process doesn't warm the cache? Any way to include that?

All subsequent SCSS saves will do a css hot reload without triggering a page refresh ✅ 

Not to let perfect be the enemy of good - as-is, this PR will greatly improve the behavior of ember-auto-import with CSS hot reloads.

packages/ember-auto-import/ts/webpack.ts
```diff
@@ -526,7 +535,10 @@ export default class WebpackBundler extends Plugin implements Bundler {
             return '[' + deps.map((d) => `'${d}'`).join(',') + ']';
           }
         );
-        writeFileSync(resolve(this.outputPath, assetFile), outputSrc, 'utf8');
```
apellerano-pw, line 529: According to the docs, `'utf8'` is the default value, so I just dropped it here to avoid more arity on `cachedWriteFileSync`

mansona (approved):
great thanks for this! 🎉

PR #684, opened 2026-01-21
fix: remove hardcoded source for hash

lifeart:
at the moment we have issue where all modules with ', " has same hash id


lifeart:
closing because it's included into https://github.com/embroider-build/ember-auto-import/pull/683

PR #711, opened 2026-05-06
Test out ember-source w/ sideEffects: false

NullVoxPopuli:
Companion PR: 
- https://github.com/emberjs/ember.js/pull/21365


NullVoxPopuli:
it works, so I close the PR, since we don't actually want to test against a custom build of ember forever

